PR #30, opened 2019-07-12
Add version number to downloaded weight file

lgeiger:
This allows us to update the weight files later and don't break user space


lgeiger:
This currently fails because of https://github.com/larq/larq/pull/136

PR #35, opened 2019-07-15
Accuracies and hparams for pretrained binary alexnet

koenhelwegen:


jamescook106:
These look good to me, but is it a draft? The PR is normal but the title is draft

larq_zoo/utils.py
```diff
@@ -11,6 +11,27 @@
 ImagenetDataset = namedtuple("ImagenetDataset", ["input_shape", "num_classes"])
 
 
+def slash_join(*args):
+    return "/".join(arg.strip("/") for arg in args)
+
+
+def download_pretrained_model(model, version, file, file_hash, cache_dir=None):
```
lgeiger: How do you feel about consolidating `file` and `version` into a single argument called `tag`?

larq_zoo/utils.py
```diff
@@ -11,6 +11,27 @@
...
+    root_url = "https://github.com/plumerai/larq-zoo/releases/download/"
+
+    url = slash_join(root_url, model + "-" + version, file)
+    cache_subdir = os.path.join("larq/models/", model)
+
+    weights_path = tf.keras.utils.get_file(
```
lgeiger: I'd just do `return tf.keras.utils.get_file(` here and remove the return below.

larq_zoo/birealnet.py
```diff
@@ -143,22 +143,19 @@ def BiRealNet(
...
+                file="birealnet_weights.h5",
                 file_hash="7fafcef8c453ec528fce3562fd5c7835",
             )
         else:
-            weights_path = tf.keras.utils.get_file(
-                "birealnet_weights_notop-v0.1.0.h5",
-                url + "birealnet_weights_notop.h5",
-                cache_dir=cache_dir,
-                cache_subdir="models",
+            weights_path = utils.download_pretrained_model(
```
lgeiger: Thanks for cleaning this up.

larq_zoo/utils.py
```diff
@@ -11,6 +11,27 @@
...
 
 
+def slash_join(*args):
+    return "/".join(arg.strip("/") for arg in args)
+
+
+def download_pretrained_model(model, version, file, file_hash, cache_dir=None):
+    root_url = "https://github.com/plumerai/larq-zoo/releases/download/"
+
+    url = slash_join(root_url, model + "-" + version, file)
```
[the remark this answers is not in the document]
koenhelwegen: I looked into this, but `urlparse.urljoin` seems to be intended to parse proper urls rather than do easy concatenating (see e.g. [this](https://stackoverflow.com/a/1793282/3425529) SO question)

larq_zoo/utils.py
koenhelwegen: doesn't work if we want to leave out the version in the locally stored copy

docs/index.md
```diff
@@ -24,3 +24,4 @@ The model definitions and the train loop are available in the [Larq Zoo reposito
 | Model                                 | Top-1 Accuracy | Top-5 Accuracy |
 | ------------------------------------- | -------------- | -------------- |
 | [Bi-Real Net](/models/api/#birealnet) | 55.88 %        | 78.62 %        |
+| [Binary Alexnet](/models/api/#BinaryAlexNet) | 36.28 % | 61.05 %        |
```
lgeiger: I am not sure if this link needs to be lower case, but let's try it like this.

PR #38, opened 2019-07-17
Add integration test for training loop

lgeiger:
This test will take around 5 minutes which isn't ideal, but I think it is a useful sanity check to have.


lgeiger:
This probably fails due to https://github.com/larq/zookeeper/pull/32

lgeiger:
This moves us about 90% code coverage 🎉

PR #54, opened 2019-08-08
Add links to view model architecture with netron

lgeiger:
This will add links to the documentation which allows us to visualize the networks with netron in the browser. For this to work I needed to use a CORS proxy to workaround the default CORS policy set by GitHub pages. See https://github.com/lutzroeder/netron/issues/312

Note that for BinaryNet and XNOR-Net it might take quite long to open the website since it needs to first download the model file which also include all the weights.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -10,7 +10,7 @@ BinaryAlexNet(include_top=True,
               classes=1000)
 ```
 
-Instantiates the Binary AlexNet architecture.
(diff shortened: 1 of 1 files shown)


lgeiger:
> This is great. It really takes a long time to open the website though. I am not sure if I would wait that long if there was no warning (like there is in the commit message).

I agree, birealnet loads in ~15s last time I tested, Alexnet and XNOR- Net is really huge, so there isn't much we can do to make it faster. I am not sure if netron supports loading only the architecture from a json file.

MariaHeuss:
What do you think about adding this warning with a start (*) on the end of the page or something similar such that people know it is supposed to take this long? I would just think something is not working properly if I did not know this and close the page again.

lgeiger:
I think it should be possible to only load the architecture and not the weights to speed up loading.

I'll do that in a new PR

PR #59, opened 2019-08-13
[Not ready for merge] Add ResNetE18 model

lgeiger:
This adds the ResNetE18 model from https://arxiv.org/abs/1906.08637

This PR will currently fail CI since it requires https://github.com/larq/larq/pull/182 to be released first as Larq 0.7.0 first.


docs/index.md
```diff
@@ -15,6 +15,7 @@ The model definitions and the train loop are available in the [Larq Zoo reposito
 | -------------------------------------------- | -------------- | -------------- | ---------- | -------- |
 | [Binary AlexNet](/models/api/#binaryalexnet) | 36.30 %        | 61.53 %        | 61 859 192 | 7.49 MB  |
 | [Bi-Real Net](/models/api/#birealnet)        | 58.01 %        | 80.27 %        | 11 736 232 | 4.04 MB  |
+| [ResNetE18](/models/api/#resnete18)          | 58.26 %        | 80.84 %        | 11 699 368 | 4.03 MB  |
 | [XNOR-Net](/models/api/#xnornet)             | 44.96 %        | 69.18 %        | 62 396 768 | 22.81 MB |
```
jamescook106: It's good to see you have beaten the paper accuracy, but do you think that it could go higher? Looking at it from this table it looks pretty much identical to Bi-RealNet in terms of accuracy, params and model size

docs/index.md
```diff
@@ -15,6 +15,7 @@ The model definitions and the train loop are available in the [Larq Zoo reposito
 | -------------------------------------------- | -------------- | -------------- | ---------- | -------- |
```
jamescook106: Should we reorder this table by accuracy?

larq_zoo/resnet_e.py
```diff
@@ -0,0 +1,182 @@
...
+    epochs = 120
+    batch_size = 1024
+    num_layers = 18
+    learning_rate = 0.004
+    learning_factor = 0.3
+    learning_steps = [70, 90, 110]
+    initial_filters = 64
+    quantizer = lq.quantizers.SteSign(clip_value=1.25)
+    constraint = lq.constraints.WeightClip(clip_value=1.25)
+
```
jamescook106: Why is the clip_value 1.25? Is this something from the supplementary material?

lgeiger: That's what they used in the paper and the [supplementary materials](https://owncloud.hpi.de/s/1jrAUnqRAfg0TXH) also explores different clipping values.

docs/index.md
lgeiger: I reran the [example](https://larq.dev/models/examples/#evaluate-bi-real-net-with-tensorflow-datasets) and updated the table with the exact accuracies (58.32 % Top-1 and 80.79 % Top-5).

I didn't spend any time tuning this network, so I think there is still room for improvement, but that's not a high priority for now. I think it is still a valuable addition since the network is conceptionally simpler than BiRealNet (just uses STE) and trains in less epochs to the claimed accuracy.

lgeiger:
Closing in favour of #60 to make rebasing easier

PR #60, opened 2019-08-17
Add ReseNetE18 and BinaryDenseNet{28,37,45}

lgeiger:
This adds the DenseNet28 and DenseNet45 models from https://arxiv.org/abs/1906.08637

~~Note: This PR is build on top of #59 so only the last two commits are relevant for review here.~~


docs/api.md
```diff
@@ -131,6 +131,50 @@ A Keras model instance.
 
 **References**
 
+- [Back to Simplicity:
+  How to Train Accurate BNNs from Scratch?](https://arxiv.org/abs/1906.08637)
+
+## DenseNet28<a class="headerlink code-link" style="float:right;" href="https://github.com/larq/zoo/blob/master/larq_zoo/densenet.py#L119" title="Source code"></a>
+
+```python
+DenseNet28(include_top=True,
```
timdebruin: To avoid ambiguity when comparing full precision architectures should we call this BinaryDensenet? Of course we could `import Densenet as BinaryDensenet` but being explicit (and calling it exactly as in the paper) might prevent some confusion.

larq_zoo/densenet.py
```diff
@@ -0,0 +1,181 @@
...
+        use_bias=False,
+    )(y)
+    return keras.layers.concatenate([x, y])
+
+
+@registry.register_model
+def densenet(args, dataset, input_tensor=None, include_top=True):
+    input = utils.get_input_layer(dataset.input_shape, input_tensor)
+
+    if dataset.input_shape[0] and dataset.input_shape[0] < 50:
```
timdebruin: Same as above

larq_zoo/densenet.py
```diff
@@ -0,0 +1,181 @@
...
+        input_quantizer=args.quantizer,
+        kernel_quantizer=args.quantizer,
+        kernel_initializer="glorot_normal",
+        kernel_constraint=args.constraint,
+        padding="same",
+        use_bias=False,
+    )(y)
+    return keras.layers.concatenate([x, y])
+
+
```
MariaHeuss: Maybe it would be good to have a comment somewhere on how the imageshape of the dataset should look like for all the models we have. Especially if there are different possible image sizes as here.

larq_zoo/densenet.py
MariaHeuss: Maybe it would be good to have a comment somewhere on how the imageshape of the dataset should look like for all the models we have. Especially if there are different possible image sizes as here.

larq_zoo/densenet.py
```diff
@@ -0,0 +1,181 @@
...
+@registry.register_hparams(densenet)
+class densenet37(densenet28):
+    batch_size = 192
+    learning_steps = [80, 100]
+    reduction = [3.3, 3.3, 4]
+    layers = [6, 8, 12, 6]
+
+
+@registry.register_hparams(densenet)
+class densenet45(densenet37):
```
MariaHeuss: would not make more sense to inherit from densenet28 again?

larq_zoo/densenet.py
```diff
@@ -0,0 +1,181 @@
...
+        kernel_initializer="glorot_normal",
+        kernel_constraint=args.constraint,
+        padding="same",
+        use_bias=False,
+    )(y)
+    return keras.layers.concatenate([x, y])
+
+
+@registry.register_model
+def densenet(args, dataset, input_tensor=None, include_top=True):
```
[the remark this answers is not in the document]
lgeiger: Rebased it 👍

larq_zoo/densenet.py
lgeiger: It is mentioned in the docstrings. This line is mostly in order to allow training with CIFAR10

docs/api.md
lgeiger: I agree. That seams like a good idea.

PR #62, opened 2019-08-19
Made compatible with Zookeeper >= 0.5

### LINKED ISSUE #57: Update Larq/Zoo to work with the breaking changes of Larq/Zookeeper
Zookeeper has been updated such that it can support a variety of computer vision tasks. However, the update to Zookeeper contained a breaking change (see https://github.com/larq/zookeeper/pull/40). Therefore before the version of Zookeeper for Zoo can be updated (see #55), the breaking changes must be incorporated

timdebruin:
This fixes #57 closes #55

### DIFF
--- a/larq_zoo/data.py
+++ b/larq_zoo/data.py
@@ -5,7 +5,8 @@ The ImageNet preprocessing from https://github.com/tensorflow/tpu/blob/master/mo
 
 import numpy as np
 import tensorflow as tf
-from zookeeper import registry
+from zookeeper import registry, Preprocessing
+import tensorflow_datasets as tfds
 
 IMAGE_SIZE = 224
 CROP_PADDING = 32
@@ -31,12 +32,34 @@ def preprocess_input(image):
     return result
 
 
-@registry.register_preprocess("imagenet2012", (IMAGE_SIZE, IMAGE_SIZE, 3))
-@registry.register_preprocess("oxford_flowers102", (IMAGE_SIZE, IMAGE_SIZE, 3))
-def default(image_bytes, training):
-    return preprocess_image_bytes(
-        image_bytes, is_training=training, image_size=IMAGE_SIZE
-    )
+class ImageClassification(Preprocessing):
+    @property
+    def kwargs(self):
+        return {
+            "input_shape": self.features["image"].shape,
+            "num_classes": self.features["label"].num_classes,
+        }
+
+    def inputs(self, data, training):
+        return tf.cast(data["image"], tf.float32)
+
+    def outputs(self, data, training):
+        return tf.one_hot(data["label"], self.features["label"].num_classes)
+
+
+@registry.register_preprocess("imagenet2012")
+@registry.register_preprocess("oxford_flowers102")
+class default(ImageClassification):
+    decoders = {"image": tfds.decode.SkipDecoding()}
+
+    @property
+    def kwargs(self):
+        return {**super().kwargs, "input_shape": (IMAGE_SIZE, IMAGE_SIZE, 3)}
+
+    def inputs(self, data, training):
+        return preprocess_image_bytes(
+            data["image"], is_training=training, image_size=IMAGE_SIZE
+        )
 
 
 def distorted_bounding_box_crop(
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -8,7 +8,7 @@ def readme():
 
 setup(
     name="larq-zoo",
-    version="0.3.0",
+    version="0.4.0",
```
timdebruin: I am not sure if this is right. Since this is breaking (in that the new zookeeper version is incompatible with the old one) I assume we go from 0.3.0 to 0.4.0.

larq_zoo/data.py
```diff
@@ -31,12 +32,34 @@ def preprocess_input(image):
...
+class ImageClassification(Preprocessing):
+    @property
+    def kwargs(self):
+        return {
+            "input_shape": self.features["image"].shape,
+            "num_classes": self.features["label"].num_classes,
+        }
+
+    def inputs(self, data, training):
+        return tf.cast(data["image"], tf.float32)
```
MariaHeuss: Why is the training argument in this function?

timdebruin: The `inputs` and `outputs` functions here overwrite those functions form the [`Preprocessing`](https://github.com/larq/zookeeper/blob/1210127673e3c4a9bd1ec1bcdc4fc6200b16797d/zookeeper/preprocessing.py#L38) class which has the `training` argument. The input function is overwritten below where it uses the argument. I included it here to keep the method signature the same for clarity.

larq_zoo/utils.py
```diff
@@ -8,8 +8,6 @@
 from keras_applications.imagenet_utils import _obtain_input_shape
 from collections import namedtuple
 
-ImagenetDataset = namedtuple("ImagenetDataset", ["input_shape", "num_classes"])
-
```
MariaHeuss: Are you sure we do not need this anymore? If so the import should also be deleted.

setup.py
```diff
@@ -17,7 +17,7 @@ def readme():
     url="https://github.com/plumerai/larq-zoo",
     packages=find_packages(),
     license="Apache 2.0",
-    install_requires=["numpy~=1.15", "larq~=0.5", "zookeeper>=0.3.1,<0.5.0"],
+    install_requires=["numpy~=1.15", "larq~=0.5", "zookeeper~=0.5"],
```
MariaHeuss: Doesn't it require zookeeper 0.5? If so I would use >=0.5.

setup.py
MariaHeuss: It looks like so far each release had either some new model or improved weights of already existing models in it. For my feeling this is more a 0.3.1 but I am also a bit uncertain about this.

larq_zoo/utils.py
timdebruin: There were no references to this anymore in larq_zoo. As far as I could tell the tuple was used as a dummy dataset with only the relevant properties of `input_shape` and `num_classes`. Since `datasets` are no longer passed to models in the same way this should not be necessary anymore. d529090cf5d1272b7aa3438f1764d9660a948975 removes the import statement.

setup.py
timdebruin: It does, I have implemented your fix in 79194172937878737ada9144a6c6dd29895061bb

setup.py
timdebruin: @koenhelwegen could you shine some light here?

koenhelwegen: There is no need to change the version number when you change the code - it's fine if the master branch is non-compatible with the latest release. Releases are used to make a specific version of the code easily available (e.g. after a bug fix or when several changes are made). This is done by going to [releases](https://github.com/larq/zoo/releases) and drafting a release. When the time is there, we make a one line PR to change the version number ([example](https://github.com/larq/zoo/commit/cbe657d7a12b7872c078337270ae0c0e1750deed)).

Once we make the release I think we may want to go to 0.4 as it's a breaking change, but below 1.0.0 there are no real rules in general as far as I know.

lgeiger: Correct. There is no need to change the version number here.

setup.py
lgeiger: We should keep the dependency fixed to `zookeeper~=0.5` to prevent us from accidentally pulling in breaking changes.
For reference `~=0.5` is equivalent to `>=0.5.0,<0.6.0`

timdebruin: Reset to `~=0.5`

setup.py
timdebruin: reset to `0.3.0`

PR #66, opened 2019-09-02
Add Dilated BinaryDenseNet37

lgeiger:
This adds a dilated version of BinaryDenseNet37 which achieves ~64.3 % top-1 accuracy.

### DIFF
--- a/larq_zoo/densenet.py
+++ b/larq_zoo/densenet.py
@@ -4,11 +4,12 @@ from larq_zoo import utils
 import larq as lq
 
 
-def densely_connected_block(x, args):
+def densely_connected_block(x, args, dilation_rate=1):
     y = keras.layers.BatchNormalization(momentum=0.9, epsilon=1e-5)(x)
     y = lq.layers.QuantConv2D(
         filters=args.growth_rate,
         kernel_size=3,
+        dilation_rate=dilation_rate,
         input_quantizer=args.quantizer,
         kernel_quantizer=args.quantizer,
         kernel_initializer="glorot_normal",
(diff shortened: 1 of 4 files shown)


larq_zoo/densenet.py
```diff
@@ -49,11 +50,12 @@ def binary_densenet(
 
     for block, layers_per_block in enumerate(args.layers):
         for _ in range(layers_per_block):
-            x = densely_connected_block(x, args)
+            x = densely_connected_block(x, args, args.dilation_rate[block])
 
         if block < len(args.layers) - 1:
             x = keras.layers.BatchNormalization(momentum=0.9, epsilon=1e-5)(x)
-            x = keras.layers.MaxPooling2D(2, strides=2)(x)
+            if args.dilation_rate[block + 1] == 1:
```
timdebruin: Should we consider the case where some but not all of the max pooling layers are replaced with dilation? (e.g. dilation_rate = [1, 1, 2, 2])

lgeiger: I tried to keep this close to the [official implementation](https://github.com/hpi-xnor/BMXNet-v2-examples/blob/master/binary_models/densenet.py#L84-L87).

I think for now it is fine to only use dilation when we don't reduce the spacial dimension in the transition block.

tests/train_test.py
```diff
@@ -1,7 +1,13 @@
+import pytest
 from click.testing import CliRunner
 from larq_zoo import train
+import tensorflow as tf
 
 
+@pytest.mark.skipif(
+    int(tf.__version__[0]) == 2,
+    reason="This currently fails on TF 2 due to https://github.com/larq/larq/pull/195.",
+)
```
timdebruin: Should we test for tf version when making the model for which tf2 fails and raise a comprehensible error preemptively?

lgeiger: I don't think that's necessary for now, since TF 2 isn't stable yet and this might be a bug in TensorFlow itself. Only the training code fails, so loading the model via the public API works correctly in this version.

We can add a proper error message once we have time to dig into https://github.com/larq/larq/pull/195

PR #73, opened 2019-09-19
Correctly pass top argument to decode_predictions

lgeiger:
Previously the `top` argument in `larq_zoo. decode_predictions` wasn't respected and always returned the Top 5 predictions.


MariaHeuss:
Can you maybe briefly write a description about what breaks when you have this bug in your code?

lgeiger:
> Can you maybe briefly write a description about what breaks when you have this bug in your code?

Updated the PR description.

PR #75, opened 2019-09-25
Add DoReFa-net

MariaHeuss:
This adds the pretrained model for DoReFa with 1 bit weights and 2 bit activations.

### DIFF
--- a/larq_zoo/dorefanet.py
+++ b/larq_zoo/dorefanet.py
@@ -6,7 +6,7 @@ from zookeeper import registry, HParams
 
 @registry.register_model
 def dorefa_net(hparams, input_shape, num_classes, input_tensor=None, include_top=True):
-    def conv_block(x, features, kernel_size, strides=1, pool=False, padding="same"):
+    def conv_block(x, filters, kernel_size, strides=1, pool=False, padding="same"):
         x = lq.layers.QuantConv2D(
             features,
             kernel_size=kernel_size,
--- a/larq_zoo/dorefanet.py
+++ b/larq_zoo/dorefanet.py
@@ -8,7 +8,7 @@ from zookeeper import registry, HParams
 def dorefa_net(hparams, input_shape, num_classes, input_tensor=None, include_top=True):
     def conv_block(x, filters, kernel_size, strides=1, pool=False, padding="same"):
         x = lq.layers.QuantConv2D(
-            features,
+            filters,
             kernel_size=kernel_size,
             strides=strides,
             padding="same",
--- a/larq_zoo/dorefanet.py
+++ b/larq_zoo/dorefanet.py
@@ -25,7 +25,6 @@ def dorefa_net(hparams, input_shape, num_classes, input_tensor=None, include_top
         return x
 
     def fully_connected_block(x, units):
-
         x = lq.layers.QuantDense(
             units,
             input_quantizer=hparams.input_quantizer,
--- a/larq_zoo/dorefanet.py
+++ b/larq_zoo/dorefanet.py
@@ -47,7 +47,7 @@ def dorefa_net(hparams, input_shape, num_classes, input_tensor=None, include_top
     out = conv_block(out, filters=256, kernel_size=5, pool=True)
     out = conv_block(out, filters=384, kernel_size=3, pool=True)
     out = conv_block(out, filters=384, kernel_size=3)
-    out = conv_block(out, filters=256, kernel_size=3, padding="valid", pool=True)
+    out = conv_block(out, filters=256, kernel_size=3, pool_padding="valid", pool=True)
 
     # classifier
     if include_top:
--- a/larq_zoo/dorefanet.py
+++ b/larq_zoo/dorefanet.py
@@ -6,7 +6,7 @@ from zookeeper import registry, HParams
 
 @registry.register_model
 def dorefa_net(hparams, input_shape, num_classes, input_tensor=None, include_top=True):
-    def conv_block(x, filters, kernel_size, strides=1, pool=False, padding="same"):
+    def conv_block(x, filters, kernel_size, strides=1, pool=False, pool_padding="same"):
         x = lq.layers.QuantConv2D(
             filters,
             kernel_size=kernel_size,
@@ -21,7 +21,7 @@ def dorefa_net(hparams, input_shape, num_classes, input_tensor=None, include_top
             x
         )
         if pool:
-            x = tf.keras.layers.MaxPool2D(pool_size=3, strides=2, padding=padding)(x)
+            x = tf.keras.layers.MaxPool2D(pool_size=3, strides=2, padding=pool_padding)(x)
         return x
 
     def fully_connected_block(x, units):
--- a/larq_zoo/dorefanet.py
+++ b/larq_zoo/dorefanet.py
@@ -24,7 +24,7 @@ def dorefa_net(hparams, input_shape, num_classes, input_tensor=None, include_top
             x = tf.keras.layers.MaxPool2D(pool_size=3, strides=2, padding=padding)(x)
         return x
 
-    def dense_block(x, units):
+    def fully_connected_block(x, units):
 
         x = lq.layers.QuantDense(
             units,
@@ -53,8 +53,8 @@ def dorefa_net(hparams, input_shape, num_classes, input_tensor=None, include_top
     # classifier
     if include_top:
         out = tf.keras.layers.Flatten()(out)
-        out = dense_block(out, units=4096)
-        out = dense_block(out, units=4096)
+        out = fully_connected_block(out, units=4096)
+        out = fully_connected_block(out, units=4096)
         out = tf.keras.layers.Activation("clip_by_value_activation")(out)
         out = tf.keras.layers.Dense(num_classes, use_bias=True)(out)
         out = tf.keras.layers.Activation("softmax")(out)
@@ -65,7 +65,7 @@ def dorefa_net(hparams, input_shape, num_classes, input_tensor=None, include_top
 @lq.utils.register_keras_custom_object
 @lq.utils.set_precision(1)
 def magnitude_aware_sign_unclipped(x):
-    r"""
+    """
     Scaled sign function with identity pseudo-gradient as used for the
     weights in the DoReFa paper. The Scale factor is calculated per layer.
     """
--- a/larq_zoo/dorefanet.py
+++ b/larq_zoo/dorefanet.py
@@ -1,0 +1,183 @@
+import tensorflow as tf
+import larq as lq
+from larq_zoo import utils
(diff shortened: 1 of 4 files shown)


lgeiger:
It looks like this requires a new release of Larq. @jamescook106 @AdamHillier Do you want to release a new version, so we can get this merged?

AdamHillier:
@lgeiger just published new release 0.7.2

MariaHeuss:
How can I make sure that the tests get ran again with the newly released version of larq?

lgeiger:
> How can I make sure that the tests get ran again with the newly released version of larq?

You can bump the required version [here](https://github.com/larq/zoo/blob/master/setup.py#L20) and push a new commit.

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,183 @@
...
+        )(x)
+        x = tf.keras.layers.BatchNormalization(scale=False, momentum=0.9, epsilon=1e-4)(
+            x
+        )
+        if pool:
+            x = tf.keras.layers.MaxPool2D(pool_size=3, strides=2, padding=padding)(x)
+        return x
+
+    def dense_block(x, units):
+
```
jneeven: Can you remove this empty line, and can you verify that model.save() does not crash? Otherwise looks good!

lgeiger:
Please also bump the required larq version [here](https://github.com/larq/zoo/blob/master/setup.py#L20) to make sure the installed version of larq includes all needed quantizers.

MariaHeuss:
I rebased the branch to have the requirements. I was not sure how to bump it otherwise..

MariaHeuss:
Oh I get it.

lgeiger:
> I rebased the branch to have the requirements. I was not sure how to bump it otherwise..

Fixed it in 35e449002442b8dbab650eadf7e1dd82c6c84d28. Don't worry about it.

MariaHeuss:
Why did the commit pass the tests before the bump? @lgeiger

lgeiger:
> Why did the commit pass the tests before the bump?

`larq~=0.7.1` is short for `larq>=0.7.1,<0.8` and therefor installs the latest matching version (in this case 0.7.2) on a fresh install, but `0.7.1` would match too so `pip` won't upgrade an existing installation. That's why we need to bump it explicitly.
For a detailed explanation checkout https://semver.org

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,183 @@
...
+def dorefa_net(hparams, input_shape, num_classes, input_tensor=None, include_top=True):
+    def conv_block(x, features, kernel_size, strides=1, pool=False, padding="same"):
+        x = lq.layers.QuantConv2D(
+            features,
+            kernel_size=kernel_size,
+            strides=strides,
+            padding="same",
+            input_quantizer=hparams.input_quantizer,
+            kernel_quantizer=hparams.kernel_quantizer,
+            kernel_constraint=None,
```
lgeiger: This is the default, I don't think we need to repeat this here.
```suggestion
```

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+@registry.register_model
+def dorefa_net(hparams, input_shape, num_classes, input_tensor=None, include_top=True):
+    def conv_block(x, features, kernel_size, strides=1, pool=False, padding="same"):
+        x = lq.layers.QuantConv2D(
+            features,
+            kernel_size=kernel_size,
+            strides=strides,
+            padding="same",
```
lgeiger: Does this intentionally not use the padding argument of the function [here](https://github.com/larq/zoo/pull/75/files#diff-5bee5daf933f45f505789d9c9a479a59R9)?

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,183 @@
...
+            x = tf.keras.layers.MaxPool2D(pool_size=3, strides=2, padding=padding)(x)
+        return x
+
+    def dense_block(x, units):
+
+        x = lq.layers.QuantDense(
+            units,
+            input_quantizer=hparams.input_quantizer,
+            kernel_quantizer=hparams.kernel_quantizer,
+            kernel_constraint=None,
```
lgeiger:
```suggestion
```

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,183 @@
+import tensorflow as tf
+import larq as lq
+from larq_zoo import utils
+from zookeeper import registry, HParams
+
+
+@registry.register_model
+def dorefa_net(hparams, input_shape, num_classes, input_tensor=None, include_top=True):
+    def conv_block(x, features, kernel_size, strides=1, pool=False, padding="same"):
```
lgeiger: I'd prefer the following to keep wording consistent:
```suggestion
    def conv_block(x, filters, kernel_size, strides=1, pool=False, padding="same"):
```

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,183 @@
...
+        x = lq.layers.QuantConv2D(
+            features,
```
lgeiger:
```suggestion
            filters,
```

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,183 @@
...
+            use_bias=False,
+        )(x)
+        x = tf.keras.layers.BatchNormalization(scale=False, momentum=0.9, epsilon=1e-4)(
+            x
+        )
+        if pool:
+            x = tf.keras.layers.MaxPool2D(pool_size=3, strides=2, padding=padding)(x)
+        return x
+
+    def dense_block(x, units):
```
lgeiger: This reminds me of a block in a densenet. What do you think about 
```suggestion
    def fully_connected_block(x, units):
```

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,183 @@
...
+    )(img_input)
+    out = conv_block(out, features=256, kernel_size=5, pool=True)
+    out = conv_block(out, features=384, kernel_size=3, pool=True)
+    out = conv_block(out, features=384, kernel_size=3)
+    out = conv_block(out, features=256, kernel_size=3, padding="valid", pool=True)
+
+    # classifier
+    if include_top:
+        out = tf.keras.layers.Flatten()(out)
+        out = dense_block(out, units=4096)
```
lgeiger:
```suggestion
        out = fully_connected_block(out, units=4096)
```

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,183 @@
...
+        out = dense_block(out, units=4096)
```
lgeiger:
```suggestion
        out = fully_connected_block(out, units=4096)
```

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,183 @@
...
+        out = tf.keras.layers.Activation("clip_by_value_activation")(out)
```
lgeiger: You could also use `Activation(clip_by_value_activation)`  here or us a [Lambda Layer](https://keras.io/layers/core/#lambda). But since this introduced some problems, with serialization some time ago, let's keep it like this to not introduce any weird bugs.

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,183 @@
...
+        out = tf.keras.layers.Dense(num_classes, use_bias=True)(out)
+        out = tf.keras.layers.Activation("softmax")(out)
+
+    return tf.keras.Model(inputs=img_input, outputs=out)
+
+
+@lq.utils.register_keras_custom_object
+@lq.utils.set_precision(1)
+def magnitude_aware_sign_unclipped(x):
+    r"""
```
lgeiger: No need for raw strings since we don't use LaTeX in the doc string
```suggestion
    """
```

larq_zoo/dorefanet.py
MariaHeuss: I like to have it to underline that there is none.

larq_zoo/dorefanet.py
MariaHeuss: Oh no... This is not intentionally and means that the valid padding in the one layer is not used. I would suggest to change it to same padding in the model and put make it use the argument. Otherwise I will have to rerun the entire experiment I fear.

lgeiger: The argument is used in the [MaxPool2D](https://github.com/larq/zoo/pull/75/files/35e449002442b8dbab650eadf7e1dd82c6c84d28#diff-5bee5daf933f45f505789d9c9a479a59R24) though.

>  I would suggest to change it to same padding in the model and put make it use the argument

Could you clarify?

MariaHeuss: Yeah, the argument should indeed be called and I am using 'valid' padding later in the model (which is ignored), so if I just make it call the argument the model and the pretrained model do not fit anymore. So I would make it call the padding argument and change the 'valid' padding in the one layer to 'same'. Is this more clear?

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,182 @@
...
+        return x
+
+    # get input
+    img_input = utils.get_input_layer(input_shape, input_tensor)
+
+    # feature extractor
+    out = tf.keras.layers.Conv2D(
+        96, kernel_size=12, strides=4, padding="valid", use_bias=True
+    )(img_input)
+    out = conv_block(out, features=256, kernel_size=5, pool=True)
```
lgeiger:
```suggestion
    out = conv_block(out, filters=256, kernel_size=5, pool=True)
```

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,182 @@
...
+    out = conv_block(out, features=384, kernel_size=3, pool=True)
```
lgeiger:
```suggestion
    out = conv_block(out, filters=384, kernel_size=3, pool=True)
```

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,182 @@
...
+    out = conv_block(out, features=384, kernel_size=3)
```
lgeiger:
```suggestion
    out = conv_block(out, filters=384, kernel_size=3)
```

larq_zoo/dorefanet.py
```diff
@@ -0,0 +1,182 @@
...
+    out = conv_block(out, features=256, kernel_size=3, padding="valid", pool=True)
```
lgeiger:
```suggestion
    out = conv_block(out, filters=256, kernel_size=3, padding="valid", pool=True)
```

jneeven:
Can you run black formatting again? (I'm kind of trying to setup an automated action for this in my spare time, but it's not usable yet)

PR #79, opened 2019-10-03
Simplify HParams dynamic properties

lgeiger:
@timdebruin Just made me aware of this syntax. I think this is a lot cleaner and more readable now.


lgeiger:
Apparently the optimizer creates some variables that messes up TF 1.13. Closing for now.

PR #92, opened 2019-12-12
Zookeeper v1.0

AdamHillier:
This is a WIP update of Larq Zoo to use the new version of Zookeeper.


AdamHillier:
As is clear from the GitHub actions, I haven't updated the tests...

AdamHillier:
This is functionally complete. As you can see in `setup.py`, it depends on larq/zookeeper#110, which has yet to be merged.

I have locally tested that building all of the models works. Locally, the tests also pass except for the one that runs the CLI, which instead returns an obscure error that I have not had the time to debug.

It's probably quite important to have reviewers very carefully check that the model-building code is identical to what's on master - that all strides/paddings/momentums/learning rates et cetera are identical and there aren't any typos that have crept in. This should be straightforward, but will be time consuming; perhaps this could be split amongst multiple reviewers?

I would also like a reviewer to verify the ImageNet validation accuracy of each of the models, if possible.

AdamHillier:
In this PR I've made the choice to restrict the values that are configurable fields vs. hard-coded values. The configurable model factory fields are by-and-large restricted to quantizers and constraints, but could be expanded to, for example, include batch norm momentums, et cetera. I'm also happy for that to happen in future PRs though.

AdamHillier:
And finally all the tests pass, and happily this PR increases test coverage by several percentage points.

AdamHillier:
All that remains is to wait for larq/zookeeper#110 to be merged and for Zookeeper to make a release; otherwise, this is ready for review.

leonoverweel:
Were there some git issues with this branch that caused all the force pushes? GitHub notifications get very confused/confusing about branches/PRs with lots of force pushing so keeping that to a minimum in the future (and by default preferring merges over force-push rebases) would be nice IMO.

AdamHillier:
> Were there some git issues with this branch that caused all the force pushes? GitHub notifications get very confused/confusing about branches/PRs with lots of force pushing so keeping that to a minimum in the future (and by default preferring merges over force-push rebases) would be nice IMO.

No issue, but it's the only way to trigger re-running tests which didn't previously fail, which I needed to do in order to get the code coverage reporting to work correctly. Probably best just to ignore a thread if the notifications are a problem tbh.

leonoverweel:
Yeah muting solves the notifications problem, but in general if anyone else is using a branch force pushing changes can be quite disruptive.

AdamHillier:
> Yeah muting solves the notifications problem, but in general if anyone else is using a branch force pushing changes can be quite disruptive.

This is a PR branch that only I'm working on so I'm afraid I can't promise stability.

jneeven:
> > Yeah muting solves the notifications problem, but in general if anyone else is using a branch force pushing changes can be quite disruptive.
> 
> This is a PR branch that only I'm working on so I'm afraid I can't promise stability.

That's true, but the unfortunate reality is that me and Tim were actually using this branch (since it was the only way to use zookeeper v1 and larq-zoo at the same time), and force pushing over the old contents kinda screwed us over. I guess the problem there is that we should have made some release or at least a separate branch though; not the force pushing per se. I now fixed this by finding the overwritten commit (which interestingly enough doesn't exist in the git tree, but does exist on GitHub), downloading the code from that point, and then uploading it to one of my own repositories. It works, but that's one ugly solution...

AdamHillier:
> That's true, but the unfortunate reality is that me and Tim were actually using this branch (since it was the only way to use zookeeper v1 and larq-zoo at the same time), and force pushing over the old contents kinda screwed us over. I guess the problem there is that we should have made some release or at least a separate branch though; not the force pushing per se. I now fixed this by finding the overwritten commit (which interestingly enough doesn't exist in the git tree, but does exist on GitHub), downloading the code from that point, and then uploading it to one of my own repositories. It works, but that's one ugly solution...

Sure, that makes sense, though again relying on a Git tag that's a branch name rather than a tag or commit ID is dangerous for this very reason. If I had made significant changes without force pushing this would also have broken your usage. Luckily, once this branch is merged, you will no longer need to rely on this branch (or your copy of it).

larq_zoo/model_factory.py
```diff
@@ -0,0 +1,45 @@
...
+
+# FIXME
+QuantizerType = Union[
+    tf.keras.layers.Layer, Callable[[tf.Tensor], tf.Tensor], str, None
+]
+ContraintType = Union[tf.keras.constraints.Constraint, str, None]
+DimType = Optional[int]
+
+
+class ModelFactory:
```
AdamHillier: Perhaps this class definition belongs in Zookeeper rather than Zoo? It seems to me like it would be useful to any Zookeeper consumer, and in larq/zookeeper#110 I deleted the existing built-in `Model` class, so this could replace it. Thoughts @ reviewers?

jneeven: Yes, as long as the number of classes is optional when you don't include the top, this seems a useful class to have :+1:

larq_zoo/model_factory.py
```diff
@@ -0,0 +1,45 @@
...
+    kernel_constraint: ContraintType = Field()
+
+    # This field is included for automatic inference of `num_clases`, if no
+    # value is otherwise provided.
+    dataset: Optional[Dataset] = ComponentField(None)
+
+    @Field
+    def num_classes(self) -> int:
+        if self.dataset is None:
+            raise TypeError("Must override either `dataset` or `num_classes`.")
```
MariaHeuss: I find this error slightly confusing. Maybe adding something like "No dataset defined. Must override..." would help?

larq_zoo/model_factory.py
MariaHeuss: Yes, I think this would be useful to have in zookeeper.

AdamHillier: I've discussed this with @lgeiger; we're going to leave it in Zoo for now (can always be imported from Zoo) but can transition it to Zookeeper later on.

larq_zoo/binary_alex_net.py
```diff
@@ -0,0 +1,174 @@
...
+
+    input_quantizer = Field("ste_sign")
+    kernel_quantizer = Field("ste_sign")
+    kernel_constraint = Field("weight_clip")
+
+    @Field
+    def num_classes(self) -> int:
+        if self.dataset is None:
+            raise TypeError("Must override either `dataset` or `num_classes`.")
+        return self.dataset.num_classes
```
MariaHeuss: Isn't this already inherited from the ModelFactory class?

AdamHillier: Thanks for spotting that!

setup.py
```diff
@@ -17,10 +17,14 @@ def readme():
     url="https://github.com/plumerai/larq-zoo",
     packages=find_packages(),
     license="Apache 2.0",
-    install_requires=["numpy~=1.15", "larq~=0.8.1", "zookeeper~=0.5.3"],
+    install_requires=[
+        "numpy~=1.18",
+        "larq~=0.8.4",
+        "zookeeper @ git+git://github.com/AdamHillier/zookeeper@factories",
```
jneeven: This makes `pip install -e .` fail: 
```
Collecting zookeeper@ git+git://github.com/AdamHillier/zookeeper@factories
  Cloning git://github.com/AdamHillier/zookeeper (to revision factories) to /tmp/pip-install-pk5wh8lv/zookeeper
  Running command git clone -q git://github.com/AdamHillier/zookeeper /tmp/pip-install-pk5wh8lv/zookeeper
  WARNING: Did not find branch or tag 'factories', assuming revision or ref.
  Running command git checkout -q factories
  error: pathspec 'factories' did not match any file(s) known to git
ERROR: Command errored out with exit status 1: git checkout -q factories Check the logs for full command output.
```

AdamHillier: Ah yes sorry that'll be because I deleted the branch after it was merged into Zookeeper master. Waiting on one last thing before I do the Zookeeper release and can update the dependency here.

AdamHillier: In the meantime I'll point this to master.

AdamHillier: Now updated to the just-released new Zookeeper version.

larq_zoo/binary_alex_net.py
AdamHillier: Removed in a4ce49b

larq_zoo/train.py
```diff
@@ -1,65 +1,132 @@
...
 
-    # Save weights without top
-    notop_model = build_model(
-        hparams, **dataset.preprocessing.kwargs, include_top=False
+    metrics: List[Union[Callable[[tf.Tensor, tf.Tensor], float], str]] = Field(
+        lambda: ["sparse_categorical_accuracy", "sparse_top_k_categorical_accuracy"]
     )
-    notop_model.set_weights(model.get_weights()[: len(notop_model.get_weights())])
-    notop_model.save_weights(path.join(output_dir, f"{model_name}_weights_notop.h5"))
+
```
jneeven:
```suggestion
learning_rate_schedule: Optional[Callable] = None
```

Since `BiRealNet` currently does not have an LR schedule and the callbacks check `if self.learning_rate_schedule`, training BRN will throw an error. Also, should we change those checks to `if self.learning_rate_schedule is not None` as with object detection?

AdamHillier: Nice, thanks :+1:

jneeven:
Just to have this info under the relevant PR: I evaluated all the experiments on ImageNet, obtaining the following accuracies:

image

larq_zoo/utils.py
```diff
@@ -34,7 +42,7 @@ def get_current_epoch(output_dir):
     try:
         with open(os.path.join(output_dir, "stats.json"), "r") as f:
             return json.load(f)["epoch"]
-    except:
+    except Exception:
         return 0
```
jneeven: Should we not raise the original Exception here? We can just replace `return 0` by `raise`, right?

larq_zoo/densenet.py
```diff
@@ -1,425 +1,459 @@
...
-            x = keras.layers.Conv2D(
-                round(x.shape.as_list()[-1] // args.reduction[block] / 32) * 32,
-                kernel_size=1,
+        )(y)
+        return tf.keras.layers.concatenate([x, y])
+
+    def build(self) -> BinaryDenseNet:
+        image_input = self.image_input
+
+        if image_input.shape[1] and image_input.shape[2] < 50:
```
MariaHeuss: See comment in Resnet

larq_zoo/train.py
```diff
@@ -1,65 +1,132 @@
...
+        with utils.get_distribution_scope(self.batch_size):
+            self.model.compile(
+                optimizer=self.optimizer, loss=self.loss, metrics=self.metrics,
+            )
+
+            lq.models.summary(self.model)
+
+            if initial_epoch > 0:
+                self.model.load_weights(self.model_path)
```
jneeven: Does this correctly restore the optimizer state? I see we don't use a checkpoint manager here

jneeven: In general, this class contains a lot of duplicate code with the internal `PolyaxonExperiment` we have. Does it make sense to add some abstraction of this to zookeeper instead?

larq_zoo/model_factory.py
```diff
@@ -0,0 +1,49 @@
...
+import tensorflow as tf
+from zookeeper import ComponentField, Field
+from zookeeper.tf import Dataset
+
+from larq_zoo import utils
+
+QuantizerType = Union[
+    tf.keras.layers.Layer, Callable[[tf.Tensor], tf.Tensor], str, None
+]
+ContraintType = Union[tf.keras.constraints.Constraint, str, None]
```
jneeven:
```suggestion
ConstraintType = Union[tf.keras.constraints.Constraint, str, None]
```

However, does it make sense to define this type at all? It's only used once...

larq_zoo/model_factory.py
```diff
@@ -0,0 +1,49 @@
...
+DimType = Optional[int]
+
+
+class ModelFactory:
+    """A base class for Larq Zoo models. Defines some common fields."""
+
+    # Don't set any defaults here.
+    input_quantizer: QuantizerType = Field()
+    kernel_quantizer: QuantizerType = Field()
+    kernel_constraint: ContraintType = Field()
```
jneeven:
```suggestion
    kernel_constraint: ConstraintType = Field()
```

larq_zoo/datasets.py
```diff
@@ -0,0 +1,30 @@
...
+
+@component
+class ImageNet(TFDSDataset):
+    name = Field("imagenet2012:5.0.*")
+    train_split = Field("train")
+    validation_split = Field("validation")
+
+
+@component
+class Cifar10(TFDSDataset):
```
jneeven: We could move all of these to zookeeper too, no?

larq_zoo/birealnet.py
```diff
@@ -1,172 +1,188 @@
...
-        out = tf.keras.layers.GlobalAvgPool2D()(out)
-        out = tf.keras.layers.Dense(num_classes, activation="softmax")(out)
+        return tf.keras.layers.add([x, shortcut])
 
-    return tf.keras.Model(inputs=img_input, outputs=out, name="birealnet18")
+    def build(self) -> tf.keras.models.Model:
+        input_shape = utils.validate_input(
+            self.input_shape, self.weights, self.include_top, self.num_classes
+        )
+        img_input = utils.get_input_layer(input_shape, self.input_tensor)
```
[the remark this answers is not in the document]
AdamHillier: Yes thank you! I noticed that one too and it's already fixed.

larq_zoo/birealnet.py
```diff
@@ -1,172 +1,188 @@
...
 
 
 def BiRealNet(
-    include_top=True,
-    weights="imagenet",
-    input_tensor=None,
-    input_shape=None,
-    classes=1000,
-):
+    *,  # Keyword arguments only
```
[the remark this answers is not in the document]
AdamHillier: If you have `def fn(a, b, *, c=5, d=3.14): ...` then basically you can't pass `c` or `d` as positional arguments, so `fn(1, 2, 3, 4)` would fail; you would have to do `fn(1, 2, c=3, d=4)`.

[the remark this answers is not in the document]
AdamHillier: It's a good way to force people to be explicit when calling the function to reduce accidental errors, e.g. TF uses this so that you can't accidentally swap `logits` and `labels` when calling their various loss functions.

larq_zoo/train.py
```diff
@@ -1,65 +1,132 @@
...
-        verbose=2 if tensorboard else 1,
-        initial_epoch=initial_epoch,
-        callbacks=callbacks,
-    )
 
-    model_name = build_model.__name__
-    model.save_weights(path.join(output_dir, f"{model_name}_weights.h5"))
+    @property
+    def model_path(self):
+        return self.output_dir / "model"
```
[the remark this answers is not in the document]
AdamHillier: Ah yes very true, thanks will fix.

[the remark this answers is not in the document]
AdamHillier: Fixed in 93a3771.

larq_zoo/model_factory.py
AdamHillier: Fixed in 93a3771.

larq_zoo/train.py
AdamHillier: Talked to @koenhelwegen about this; the training loop is much less important than the models, and we can push any changes that aren't bug fixes to future PRs.

larq_zoo/utils.py
AdamHillier: We want to return zero, because in particular an exception would be raised if the file doesn't exist, as is the case at the start of training.

larq_zoo/train.py
AdamHillier: I believe that you might be right about this not restoring optimizer state (at least not for multi-GPU) but as this was the previous implementation (https://github.com/larq/zoo/blob/2e62a7a0dc266cbaf4d56a1d1e95703b54eeb06a/larq_zoo/train.py#L39-L41) I think we can leave it for now. Fixing that can be pushed to a future PR.

larq_zoo/datasets.py
AdamHillier: I do broadly agree with you on this, but I think it falls into the same bucket as the `ModelFactory` class - something that we can maybe move later.

larq_zoo/train.py
```diff
@@ -1,65 +1,132 @@
...
 
-    # Save weights without top
-    notop_model = build_model(
-        hparams, **dataset.preprocessing.kwargs, include_top=False
+    metrics: List[Union[Callable[[tf.Tensor, tf.Tensor], float], str]] = Field(
+        lambda: ["sparse_categorical_accuracy", "sparse_top_k_categorical_accuracy"]
```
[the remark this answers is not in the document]
AdamHillier: Yes I believe so.

larq_zoo/dorefanet.py
```diff
@@ -84,46 +30,109 @@ def clip_by_value_activation(x):
...
-    include_top=True,
-    weights="imagenet",
-    input_tensor=None,
-    input_shape=None,
-    classes=1000,
-):
+    *,  # Keyword arguments only
+    input_shape: Tuple[int, int, int] = None,
+    input_tensor: tf.Tensor = None,
+    weights: str = "imagenet",
```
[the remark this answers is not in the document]
AdamHillier: Very good point, thank you.

larq_zoo/train.py
koenhelwegen: definitely wouldn't worry about multi-gpu training or restoring. The training loops are here so people can see what we have done, meaning simplicity should be prioritized over supporting a range of complex infrastructure options.

larq_zoo/densenet.py
AdamHillier: Resolved by db4b68c.

larq_zoo/resnet_e.py
```diff
@@ -112,28 +31,128 @@ def spec(self):
...
-    include_top=True,
-    weights="imagenet",
-    input_tensor=None,
-    input_shape=None,
-    classes=1000,
-):
+    *,  # Keyword arguments only
+    input_shape: Tuple[int, int, int] = None,
+    input_tensor: tf.Tensor = None,
+    weights: str = "imagenet",
```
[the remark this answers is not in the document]
AdamHillier: Resolved by db4b68c.

larq_zoo/dorefanet.py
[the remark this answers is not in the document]
AdamHillier: Resolved by db4b68c.

larq_zoo/resnet_e.py
```diff
@@ -112,28 +31,128 @@ def spec(self):
...
+            metrics=[],
+        )(x)
+        x = tf.keras.layers.BatchNormalization(momentum=0.9, epsilon=1e-5)(x)
+
+        return tf.keras.layers.add([x, residual])
+
+    def build(self) -> tf.keras.models.Model:
+        image_input = self.image_input
+
+        if image_input.shape[1] and image_input.shape[2] < 50:
```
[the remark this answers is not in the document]
AdamHillier: The input shape can be e.g. `[None, None, None, 3]` and loading models will still work as long as `include_top` is `False`. None of the convolutional layers need the height/width to be defined in order for the model to build.

larq_zoo/densenet.py
AdamHillier: Same answer as above (although I fixed a typo on this line)

larq_zoo/birealnet.py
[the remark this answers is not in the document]
MariaHeuss: Intresting I did not know this yet.

larq_zoo/resnet_e.py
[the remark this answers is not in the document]
MariaHeuss: Alright I see the necessity then but there is still the typo in this line I think.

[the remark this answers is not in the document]
AdamHillier: What's the typo that's still there? Could you comment with a suggested change?

larq_zoo/resnet_e.py
```diff
@@ -112,28 +31,128 @@ def spec(self):
...
+            kernel_initializer="glorot_normal",
+            use_bias=False,
+            metrics=[],
+        )(x)
+        x = tf.keras.layers.BatchNormalization(momentum=0.9, epsilon=1e-5)(x)
+
+        return tf.keras.layers.add([x, residual])
+
+    def build(self) -> tf.keras.models.Model:
+        if self.image_input.shape[1] and self.image_input.shape[2] < 50:
```
MariaHeuss:
```suggestion
        if self.image_input.shape[1] and self.image_input.shape[1] < 50:
```

larq_zoo/resnet_e.py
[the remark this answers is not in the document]
MariaHeuss: I commented the change in the not outdated code again. It is the same typo that you corrected in the last commit for densenet

larq_zoo/utils.py
jneeven: Ah my bad, I didn't realise that it returns the latest epoch :+1:

PR #106, opened 2020-02-21
Add snapshots for model summaries

jneeven:
TODO:
 - [x] Try this directly from zookeeper without `subprocess`


tests/models_test.py
```diff
@@ -88,3 +91,28 @@ def test_no_top_variable_shape_4(app, last_feature_dim):
     input_shape = (None, None, 4)
     model = app(weights=None, include_top=False, input_shape=input_shape)
     assert model.output_shape == (None, None, None, last_feature_dim)
+
+
+@pytest.mark.parametrize("command_name", cli.commands.keys())
+def test_compare_model_summary(command_name: str, snapshot):
+    # FIXME: ideally, we'd directly call the relevant command here rather than
```
lgeiger: I agree, we definitely should just call the models directly here.

jneeven: If we only want the model summaries it'd be as easy as 
```python
@parametrize
def test_model_summary(app, last_feature_dim, capsys, snapshot):
    input_tensor = keras.layers.Input(shape=(224, 224, 3))
    model = app(weights=None, input_tensor=input_tensor)
    lq.models.summary(model)
    out, err = capsys.readouterr()
    snapshot.assert_match(out)
```

But I think it would be nice to actually compile all experiments

AdamHillier: All the models do get compiled in other tests so I think it's okay to not do that here.

jneeven: Yeah I don't mean just the models, I mean actually running the experiment (with a dry run). As far as I can see that currently doesn't happen. Though it might be worth separating the model summary snapshot and the compilation test into two separate tests

AdamHillier: Yeah I think separating the two is fine, but also with Larq Zoo the training loops are definitely less of a priority than the models + weights.

jneeven:
Rip in TFDS, I'm gonna try to get that generic dummy data up and running

jneeven:
@AdamHillier I know this is my PR, but could you update this to use #119 ? I can also do it myself, probably just not today

leonoverweel:
I don't think I like having the model summary strings in `tests/snapshots/snap_models_test.py`. If we put them all in separate files (named by their models), it'll be easier to track model changes over time and find/link to individual model summaries.

jneeven:
> I don't think I like having the model summary strings in `tests/snapshots/snap_models_test.py`. If we put them all in separate `.txt` files (named by their models), it'll be easier to track model changes over time and find/link to individual model summaries.

This is not a choice I've made; it's just how snapshots work. I'm not sure if what you suggest is possible without creating an explicit python test file for every model separately...

leonoverweel:
Ah, I see. Since we're literally just comparing the string output of `model.summary()` to the text contents of a `.txt` file though, it shouldn't be too hard to implement without the snapshottest framework right?

jneeven:
> Ah, I see. Since we're literally just comparing the string output of `model.summary()` to the text contents of a `.txt` file though, it shouldn't be too hard to implement without the snapshottest framework right?

Yes true, I think I like this suggestion

AdamHillier:
The nice thing about the snapshot module is that there is a single pytest command to update the snapshot. As long as there is a nice way to replicate that with a single command, I'm happy to have individual txt files.

jneeven:
> The nice thing about the snapshot module is that there is a single pytest command to update the snapshot. As long as there is a nice way to replicate that with a single command, I'm happy to have individual txt files.

As long as we keep these summaries in a specific folder, we can just delete the folder and re-generate the snapshots that way?

AdamHillier:
> As long as we keep these summaries in a specific folder, we can just delete the folder and re-generate the snapshots that way?

Sure, if that's how it works then that sounds good.

tests/models_test.py
```diff
@@ -90,3 +93,38 @@ def test_no_top_variable_shape_4(app, last_feature_dim):
...
+    else:
+        with open(summary_file, "w") as file:
+            file.write(out)
+
+
+@pytest.mark.parametrize("command_name", cli.commands.keys())
+def test_experiments(command_name: str, snapshot, capsys):
+    try:
+        cli.commands[command_name](
+            ["dataset=DummyOxfordFlowers", "batch_size=8", "testing=True"]
```
AdamHillier: The size of the dataset is 4, so maybe set this batch size to 4 (or maybe 2)?

larq_zoo/train.py
```diff
@@ -25,6 +25,9 @@ class TrainLarqZooModel(Experiment):
     # Use a per-batch progress bar (as opposed to per-epoch).
     use_progress_bar: bool = Field(False)
 
+    # Whether this experiment is compilation-only (i.e. no training)
+    testing: bool = Field(False)
```
AdamHillier: Can we rename this to `dry_run` if it's necessary?

tests/models_test.py
AdamHillier: Also can we get away with `epochs=0` here?

AdamHillier:
I like this, once the isort linting is fixed should be good to go.

leonoverweel:
😍 love this!

tests/models_test.py
```diff
@@ -90,3 +93,38 @@ def test_no_top_variable_shape_4(app, last_feature_dim):
...
+        / f"{app.__name__}_{last_feature_dim}.txt"
+    )
+
+    if summary_file.exists():
+        with open(summary_file, "r") as file:
+            content = file.read()
+        assert content == out
+    else:
+        with open(summary_file, "w") as file:
+            file.write(out)
```
AdamHillier: I could be wrong, but does this mean that if the summary file doesn't exist then when the test runs on CI it will write the .txt file (because it's missing) and the test will pass? Maybe we should raise an error here or something after writing to force test failure?

jneeven: Yes, I like that suggestion. I couldn't really think of a way to properly indicate that a new snapshot was generated, but indeed raising an error after writing the summary would be nice. Then if that was intentional, you can just re-run the tests and they'll pass :+1:

AdamHillier: Yes the warning message can be something along the lines of "Couldn't find snaphot, generating new summary, if this was intentional run the tests again and they will pass"

jneeven:
@larq/core For some reason, `keras.backend.clear_session()` (called in `parametrize`) no longer seems to have any effect; the second model to be tested will have layer names like `input_2` etc. Does anyone have any idea what might be the issue here? I'm on TF 2.0.0 and manually calling `clear_session()` makes no difference...

leonoverweel:
It also looks like it's failing on `fixture 'snapshot' not found`. Maybe some requirements/imports got messed up when you merged master?

jneeven:
> It also looks like it's failing on `fixture 'snapshot' not found`. Maybe some requirements/imports got messed up when you merged master?

Yeah that's indeed interesting, though `snapshottest` *is* still in the requirements and `pytest` is imported so I don't see why that wouldn't work...

tests/models_test.py
```diff
@@ -91,3 +94,42 @@ def test_no_top_variable_shape_4(app, last_feature_dim):
...
     model = app(weights=None, include_top=False, input_shape=input_shape)
     assert model.output_shape == (None, None, None, last_feature_dim)
+
+
+@parametrize
+def test_model_summary(app, last_feature_dim, capsys, snapshot):
+    input_tensor = keras.layers.Input(shape=(224, 224, 3))
+    model = app(weights=None, input_tensor=input_tensor)
+    lq.models.summary(model)
+    out, err = capsys.readouterr()
```
lgeiger, line 104: I'd prefer to snapshot `model.to_json()` here since this will be invariant on non-functional changes and will work across different larq versions.

lgeiger, line 104: `lq.models.summary()`  has a `print_fn` which I think should be used instead of capturing the stout which can be really flaky.

lgeiger:
Looks like you are missing the `snapshottest` pip dependency.

jneeven:
> Looks like you are missing the `snapshottest` pip dependency.

Oh wow yeah, I was looking at the wrong requirements file (zoo doesn't even have one).  I doubt that'll fix the layer name issues though

tests/models_test.py
jneeven: This would be nice for testing purposes as we're not affected by changes in formatting, but does make the files unreadable. For now I do think JSON is the best solution, if we explicitly want to have the formatted summaries somewhere we should probably create a script that adds them to the docs.

AdamHillier:
And also the `files changed` view shows way more than just this PR, presumably because of the merge commit. @leonoverweel I was under the impression that GitHub was more intelligent about that? In any case, doing a rebase and force pushing will likely help isolate the problem and see if it was a problem with the merge.

jneeven:
> And also the `files changed` view shows way more than just this PR, presumably because of the merge commit. @leonoverweel I was under the impression that GitHub was more intelligent about that? In any case, doing a rebase and force pushing will likely help isolate the problem and see if it was a problem with the merge.

It does not, @leonoverweel mentioned this early and it seems to just be a feature of the notifications overview (it shows all changes since you last checked). For me, it shows only the changes I've made in this branch.

leonoverweel:
@AdamHillier I had the same thought and messaged @jneeven about whether he merged it in a weird way maybe but... I think it's an error on our end: when you click a notification for a PR, GitHub shows a diff between the last time you viewed and the last commit on the branch, not between master and the last commit on the branch. So we were viewing the relative changes that showed master being merged; if you refresh the PR and just click to the diff view, it should show up as normal.

AdamHillier:
Ah okay, I see it now, makes sense :+1:

jneeven:
I switched the summaries to JSON and added the `snapshottest` dependency. For some reason, the binarydensenet tests take ages to pass though... Also pytest doesn't like the summaries locally, but let's see what happens here.

jneeven:
The diffs are pretty useless, but I think the fail now is just caused by the layer names :+1:

leonoverweel:
What do the `.json` diffs look like? I quite like the `.txt` ones.

leonoverweel:
Since we now have model summaries in the docs, I'm happy with using the JSON files or even going back to the single-file-with-strings approach from snapshottest. @jneeven whichever is easiest to get tests to pass again, I'm happy to approve.

jneeven:
This is pretty much done now, but tests keep failing because the `keras.backend.clear_session()` does not work properly. Not sure what to do about it...

tests/models_test.py
```diff
@@ -96,25 +96,33 @@ def test_no_top_variable_shape_4(app, last_feature_dim):
...
-    json = str(model.to_json())
+
+    class PrintToVariable:
+        output = ""
+
+        def __call__(self, x):
+            self.output += f"{x}\n"
+
+    capture = PrintToVariable()
+    lq.models.summary(model, print_fn=capture)
```
leonoverweel, line 116:
```suggestion
    import io
    from contextlib import redirect_stdout

    capture = io.StringIO()
    with redirect_stdout(f):
        lq.models.summary(model)
```

You can then use `capture.getvalue()`.

I want to refactor `lq.models.summary()` to remove the `print_fn` function anyway since I think it's a bit of an antipattern when we can use a built-in wrapper like this to achieve the same result.

tests/models_test.py
```diff
@@ -96,25 +96,33 @@ def test_no_top_variable_shape_4(app, last_feature_dim):
...
         / "model_summaries"
-        / f"{app.__name__}_{last_feature_dim}.json"
+        / f"{app.__name__}_{last_feature_dim}.txt"
     )
 
     if summary_file.exists():
         with open(summary_file, "r") as file:
             content = file.read()
-        assert content == json
+        assert content == capture.output
```
leonoverweel, line 128:
```suggestion
        assert content == capture.get_value()
```

tests/models_test.py
```diff
@@ -96,25 +96,33 @@ def test_no_top_variable_shape_4(app, last_feature_dim):
...
     else:
         with open(summary_file, "w") as file:
-            file.write(json)
+            file.write(capture.output)
```
leonoverweel, line 131:
```suggestion
            file.write(capture.get_value())
```

tests/models_test.py
leonoverweel: Never mind, using the `print_fn` is actually better because the `redirect_stdout` would break if we have other random (TF) prints happening at some points.

leonoverweel:
I'm happy to merge this once CI is green. @lgeiger mentioned we may need to clear the Keras session for this to work reliably.

jneeven:
> I'm happy to merge this once CI is green. @lgeiger mentioned we may need to clear the Keras session for this to work reliably.

This should already happen:
```python
def keras_test(func):
    """Function wrapper to clean up after TensorFlow tests.
    # Arguments
        func: test function to clean up after.
    # Returns
        A function wrapping the input function.
    """
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        output = func(*args, **kwargs)
        keras.backend.clear_session()
        return output
    return wrapper
```

Adding another `clear_session` to the *start* of the function also makes no difference.

AdamHillier:
@jneeven does this break for you locally? If so, is there a difference between running Pytest multithreaded `-n auto` and just normal single threaded?

jneeven:
> @jneeven does this break for you locally? If so, is there a difference between running Pytest multithreaded `-n auto` and just normal single threaded?

It also breaks locally, single threaded. Haven't tried multi-threaded but I guess that's exactly what CI is doing

AdamHillier:
Hopefully CI for this should pass now that #166 / #168 are merged.

AdamHillier:
There's surely something wrong with 018ed29, Larq Zoo shouldn't be on its own line should it?

jneeven:
> There's surely something wrong with [018ed29](https://github.com/larq/zoo/commit/018ed29f4a09d36fc137138e77ee07438cd82778), Larq Zoo shouldn't be on its own line should it?

Why not? It's not a third-party library in this case

AdamHillier:
> Why not? It's not a third-party library in this case

Sorry you're absolutely right, that was a dumb moment on my part 😂

leonoverweel:
So this looks like it's still failing now because auto-named layers have the wrong indices because previously-built models aren't properly being cleared out? So same problem as before :(

AdamHillier:
Haha so it turns out that this code actually works just fine, it's just that the snapshots must have been generated when it *wasn't* working because the layer names in the `.txt` files are definitely wrong, they just need to be re-generated :)

I'm about to push a commit that does that.

AdamHillier:
Well that's progress, but unfortunately it looks like TF 2.2 changes a default layer name, `tf_op_layer_Mul` -> `tf_op_layer_mul`. Not sure how we can get around that....

jneeven:
> Well that's progress, but unfortunately it looks like TF 2.2 changes a default layer name, `tf_op_layer_Mul` -> `tf_op_layer_mul`. Not sure how we can get around that without setting layer names explicitly....

I compare the summaries line by line, so we *could* just use `assert snapshot_line.lower() == current_line.lower()`... It's not extremely pretty but I don't think it's necessarily a bad solution either; in general this makes the snapshots less brittle (I don't foresee any scenario in which this would cause issues)

leonoverweel:
Nice, excited that this is ready now!

jneeven:
@lgeiger  I think this should be good to go, could you review this another time? Merging is currently blocked.

jneeven:
@larq/core do we still want this? If so, I'll resolve the conflicts and we can get this merged. If not, let's close this PR.

AdamHillier:
> @larq/core do we still want this? If so, I'll resolve the conflicts and we can get this merged. If not, let's close this PR.

I don't have a strong desire for this, and it seems like it'd be a bunch of work to get it working, so I'd be minded to close it.

lgeiger:
> @larq/core do we still want this? If so, I'll resolve the conflicts and we can get this merged. If not, let's close this PR.

I'm fairly neutral to this, although I fear that these tests might be tricky to maintain accross the multiple versions of TensorFlow which we support since the snapshots might be slightly different depending on the TensorFlow version used to generate them.

jneeven:
Would it make sense to just compare the ModelProfiles instead? Then we don't need to care about the layer names etc, but will at least be notified if e.g, the number of MACs suddenly changes unintentionally

koenhelwegen:
> Would it make sense to just compare the ModelProfiles instead? Then we don't need to care about the layer names etc, but will at least be notified if e.g, the number of MACs suddenly changes unintentionally

Would this catch issues that are not already covered by the unit tests for the `ModeProfile` itself? (https://github.com/larq/larq/blob/master/larq/models_test.py#L66)

jneeven:
> Would this catch issues that are not already covered by the unit tests for the `ModeProfile` itself? (https://github.com/larq/larq/blob/master/larq/models_test.py#L66)

It would do that check for all the models, rather than just for the toy model used in the unit tests. That test is a good way to check the calculations made in ModelProfile are still correct, but the idea behind the tests here is that we'd want to catch subtle things like accidentally enabling / disabling biases somewhere while updating some code

jneeven:
We decided that at this time, it's not worth the extra effort to change to a ModelProfile snapshot solution. I'll close this PR for now, but we can re-open it in the future if we do decide this is necessary.

PR #108, opened 2020-02-21
Split zoo into `literature` and `sota` submodule

koenhelwegen:
The `literature` submodule will contain replications from research papers (all current models). These models are intended to provide a stable reference for ideas presented in specific papers. The model implementations will be maintained, but we will not attempt to improve these models over time by applying new training strategies.

The `sota` submodule will contain top models for various scenarios. These models are intended to use in a [`SW 2.0`](https://medium.com/@karpathy/software-2-0-a64152b37c35)-like fashion. We will do our best to continuously improve the models, meaning their weights and even details about there architecture may change from release to release.

This is a breaking change.


larq_zoo/literature/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    BinaryDenseNet28,
+    BinaryDenseNet37,
+    BinaryDenseNet37Dilated,
+    BinaryDenseNet45,
+)
+from larq_zoo.literature.dorefanet import DoReFaNet
+from larq_zoo.literature.resnet_e import BinaryResNetE18
+from larq_zoo.literature.xnornet import XNORNet
+
+__all__ = [
```
jneeven: General question: does specifying `__all__` here actually do anything? It's literally a duplicated form of the code above and I *think* if you remove it you'll end up with the exact same functionality? At least e.g. `XNORNet` will be exposed this way. Though I'm not sure if specifying `__all__` here prevents you from manually accessing e.g. `larq_zoo.literature.xnornet.TrainXNORNet`?

jneeven:
Also, does it make sense to explicitly have to import each model from the relevant submodule? If at some point a network (inevitably) moves from `sota` into `literature`, that would become a breaking change. Why not just keep all models directly importable from `larq_zoo`?

koenhelwegen:
> Also, does it make sense to explicitly have to import each model from the relevant submodule? If at some point a network (inevitably) moves from `sota` into `literature`, that would become a breaking change. Why not just keep all models directly importable from `larq_zoo`?

So the idea is that the sota `quicknet` in a year may be very different from `quicknet` today - so you do actually want to clearly distinguish between the two submodules. If innovations get published in academic papers, we'll provide an implementation for that specific paper in `literature`, probably under a different name. But yes, it is a good point that this is a breaking change!

leonoverweel:
Should we add some form of what you wrote above to the README?

README.md
```diff
@@ -2,7 +2,11 @@
...
 
-Larq Zoo contains reference implementations of popular Binarized Neural Networks.
+Larq Zoo consists of a `literature` and a `sota` submodule.
+
+The `literature` submodule contains replications from research papers (all current models). These models are intended to provide a stable reference for ideas presented in specific papers. The model implementations will be maintained, but we will not attempt to improve these models over time by applying new training strategies or architecture innovations.
```
leonoverweel: Small thing but I prefer having `.md` files formatted with each sentence on its own line (over a whole paragraph on one line or hard-wrapping after `n` characters), which makes it easier to keep track of changes in git.

larq_zoo/literature/__init__.py
leonoverweel: For your last point: you can't stop people from manually accessing anything in Python really.

AdamHillier: `__all__` controls what happens if you do `from larq_zoo import *`. I think editors/IDEs might also use it.

larq_zoo/__init__.py
```diff
@@ -1,29 +1,7 @@
...
-    "BinaryDenseNet37",
-    "BinaryDenseNet37Dilated",
-    "BinaryDenseNet45",
-    "DoReFaNet",
-    "XNORNet",
-    "datasets",
-    "decode_predictions",
-    "preprocess_input",
-]
+__all__ = ["datasets", "decode_predictions", "preprocess_input", "literature", "sota"]
```
lgeiger: I think before we move to v1.0 stable, we should find a better place for `datasets`, `decode_predictions` and `preprocess_input` because they might be different for different models.

larq_zoo/literature/__init__.py
jneeven: > `__all__` controls what happens if you do `from larq_zoo import *`. I think editors/IDEs might also use it.

I see, thanks for the clarification. Indeed if I print `globals()` without specifying `__all__`, a lot more shows up. A shame that Python doesn't automatically recognize this as limiting imports...

PR #118, opened 2020-02-26
Bump version to 1.0.b2

lgeiger:


jamescook106:
I think some of the commits are missing from the drafted release notes?

lgeiger:
> I think some of the commits are missing from the drafted release notes?

Release notes will be regenerated again, once we do a release

PR #128, opened 2020-03-06
Use AveragePooling2D instead of GlobalPooling2D

### LINKED ISSUE #126: QuickNet(Large) models don't match released h5 files
### Describe the bug
The QuickNet and QuickNetLarge models that are released as h5 files on this repo and used for [the LCE benchmarks](https://docs.larq.dev/compute-engine/) do not match the architectures described in their code. The h5 files use average pool and reshape rather than GlobalAveragePool, see [this comment](https://github.com/larq/compute-engine/pull/240#issuecomment-594473667). 

The model architectures should be changed to correspond to the uploaded files, which will significantly speed up their inference time (currently, QuickNetLarge constructed from `larq-zoo` is roughly twice as slow as its h5 counterpart).

### To Reproduce
Simply construct a QuickNet model from `larq-zoo` and compare it to the released h5 models.

koenhelwegen:
AveragePooling2D has been found to be significantly faster than GlobalPooling2D when running models on LCE/TFLite.

This PR replaces global pooling for all models, not just QuickNet, to make it easier for users to build fast, LCE-compatible models.

(Note that ideally, this would be handled by TFLite or the LCE converter instead; this is basically a temporary fix.)

closes #126

### DIFF
--- a/larq_zoo/core/utils.py
+++ b/larq_zoo/core/utils.py
@@ -5,14 +5,13 @@ import sys
 from typing import Optional
 
 import tensorflow as tf
+from keras_applications.imagenet_utils import _obtain_input_shape
 from tensorflow import keras
 from tensorflow.keras.applications.vgg16 import (
     decode_predictions as keras_decode_predictions,
 )
 from tensorflow.python.eager.context import num_gpus
 from tensorflow.python.keras.backend import is_keras_tensor
-
-from keras_applications.imagenet_utils import _obtain_input_shape
 
 
 def slash_join(*args):
--- a/larq_zoo/core/utils.py
+++ b/larq_zoo/core/utils.py
@@ -103,6 +103,41 @@ def get_input_layer(input_shape, input_tensor):
     return input_tensor
 
 
+def global_pool(x: tf.Tensor, data_format: str = "channels_last") -> tf.Tensor:
+    """Global average 2D pooling and flattening.
+
+    Alternative to existing keras implementation of GlobalPooling2D.
+    AveragePooling2D is much faster than GlobalPooling2D on Larq Compute Engine.
+
+    # Arguments
+    x: 4D TensorFlow tensor.
+    data_format: data_format: A string, one of channels_last (default) or
+        channels_first. The ordering of the dimensions in the inputs. channels_last
+        corresponds to inputs with shape (batch, height, width, channels) while
+        channels_first corresponds to inputs with shape (batch, channels, height,
+        width). It defaults to "channels_last".
+
+    # Returns
+    2D TensorFlow tensor.
+
+    # Raises
+    ValueError: if tensor is not 4D or data_format is not recognized.
+    """
+    if len(x.get_shape()) != 4:
+        raise ValueError("Tensor is not 4D.")
+    if data_format not in ["channels_last", "channels_first"]:
+        raise ValueError("data_format not recognized.")
+
+    input_shape = x.get_shape().as_list()
+    pool_size = input_shape[1:3] if data_format == "channels_last" else input_shape[2:4]
+
+    x = tf.keras.layers.AveragePooling2D(pool_size=pool_size, data_format=data_format)(
+        x
+    )
+
+    return tf.keras.layers.Flatten()(x)
+
+
 def decode_predictions(preds, top=5, **kwargs):
     """Decodes the prediction of an ImageNet model.
 
--- a/larq_zoo/core/utils.py
+++ b/larq_zoo/core/utils.py
@@ -108,14 +108,14 @@ def global_pool(
 ) -> tf.Tensor:
     """Global average 2D pooling and flattening.
 
-    Alternative to existing keras implementation of GlobalAveragePooling2D.
+    Alternative to existing Keras implementation of GlobalAveragePooling2D.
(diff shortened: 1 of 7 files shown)


AdamHillier:
@koenhelwegen This doesn't work because we allow input shapes such as `(None, None, 3)` (like Keras applications) which accept images of varying spacial dimension. If we don't have knwon spacial dimension then we can't build the `AveragePool` op.

larq_zoo/core/utils.py
```diff
@@ -103,6 +103,41 @@ def get_input_layer(input_shape, input_tensor):
...
+
+    # Raises
+    ValueError: if tensor is not 4D or data_format is not recognized.
+    """
+    if len(x.get_shape()) != 4:
+        raise ValueError("Tensor is not 4D.")
+    if data_format not in ["channels_last", "channels_first"]:
+        raise ValueError("data_format not recognized.")
+
+    input_shape = x.get_shape().as_list()
```
AdamHillier: We could in theory check here if the height and width have known, static shape, and if so use the `AveragePooling2D`, otherwise use `GlobalAveragePooling2D` if one of them is `None`.

koenhelwegen: Good point.. I added a fix, but not sure if we want to merge this as it's getting rather complicated. @Tombana @lgeiger do you know how easy it would be to fix this at the converter?

Tombana: So *in theory* tflite has support for resizing the input tensor at runtime (i.e. resize from one inference pass to the next). This would not work well with the averagepool operator. However, since MobileNet also uses averagepool, I'm assuming they expect people to not resize their network at runtime. Therefore I think we can safely add such a step to the converter.

For reference:
The layer `GlobalAveragePooling2D` gets transformed to the tflite `Mean` operator which for float32 only has a reference implementation. They do seem to have [an optimized implementation](https://github.com/tensorflow/tensorflow/blob/5baff53e4e4507e4ae72ad2b1fae6bf750e5d808/tensorflow/lite/kernels/reduce.cc#L291) for int8 though!
The layer `AveragePooling2D` gets transformed to `AveragePool2D`, which has [an optimized implementation](https://github.com/tensorflow/tensorflow/blob/5baff53e4e4507e4ae72ad2b1fae6bf750e5d808/tensorflow/lite/kernels/pooling.cc#L113) for float.

AdamHillier: I think the issue is that this really should be something TFLite does automatically, it's not in the scope of LCE.

larq_zoo/core/utils.py
```diff
@@ -103,6 +103,56 @@ def get_input_layer(input_shape, input_tensor):
...
+    pool_size = input_shape[1:3] if data_format == "channels_last" else input_shape[2:4]
+
+    if not (pool_size[0] is None or pool_size[1] is None):
+
```
AdamHillier:
```suggestion
```

larq_zoo/core/utils.py
Tombana: I agree with @AdamHillier . However, for now we want a fix, and I think a temporary fix in the MLIR converter is cleaner than a temporary fix in Larq. (Assuming we could get it working properly in MLIR, that is)

larq_zoo/core/utils.py
AdamHillier: Oops sorry this appears to be something ugly that black forces

larq_zoo/core/utils.py
koenhelwegen: Yes if we could do this in the converter it would be great! Will leave the PR open for a bit in case unexpected difficulties come up.

lgeiger: I really don't think this has a place in the converter. It should either be handled by TensorFlow, or in user space.

@koenhelwegen did you try to get the shape via `tf.shape(x)` which should get the dynamic shape of the tensor? Also this is only a problem for us in zoo since we explicitely allow for `(None, None, 3)` inputs to support different input sizes.

koenhelwegen: `pool_size` only accepts ints, so `AveragePooling2D` won't work for models with dynamic input sizes.

I think for now the best solution is to put it here. I've simplified the implementation a bit and added a (minimalistic) unit test. Would be happy to merge this now.

larq_zoo/core/utils.py
```diff
@@ -103,6 +103,53 @@ def get_input_layer(input_shape, input_tensor):
...
+        )
+        x = tf.keras.layers.AveragePooling2D(
+            pool_size=pool_size, data_format=data_format
+        )(x)
+        x = tf.keras.layers.Flatten()(x)
+    except ValueError:
+        x = tf.keras.layers.GlobalAveragePooling2D(data_format=data_format, name=name)(
+            x
+        )
+
```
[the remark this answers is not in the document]
koenhelwegen: sort of, but the solution is ugly and this is an extreme edge case (someone switching between fixed and dynamic input resolution while accessing layers by their ID, which is bad practice anyway) that should be pretty easy to detect. We could put the `Flatten()` outside (vacuously applying it after `GlobalAveragePooling2D` as well) but I personally have a slight preference for keeping it like this.

PR #129, opened 2020-03-09
Make min_size in validate_input an argument for Cifar training

MariaHeuss:
The ResnetE18 model code allows it to train on small datasets such as Cifar10. Because the min_size is hardcoded the check for a valid input shape throws an error regardless. I suggest making min_size an argument such that when training on small datasets we can change the min_size. 
```   
model = ComponentField(BinaryResNetE18Factory, weights=None,, input_shape=(32, 32, 3), min_size=32)
```

### DIFF
--- a/larq_zoo/core/model_factory.py
+++ b/larq_zoo/core/model_factory.py
@@ -41,7 +41,7 @@ class ModelFactory:
 
     input_shape: Optional[Tuple[DimType, DimType, DimType]] = Field(None)
     input_tensor: Optional[tf.Tensor] = Field(None)
-    min_size: Optional[int] = Field(64)
+    min_input_resolution: Optional[int] = Field(64)
 
     @property
     def image_input(self) -> tf.Tensor:
@@ -51,7 +51,7 @@ class ModelFactory:
                 self.weights,
                 self.include_top,
                 self.num_classes,
-                self.min_size,
+                self.min_input_resolution,
             )
             self._image_input = utils.get_input_layer(input_shape, self.input_tensor)
         return self._image_input
--- a/larq_zoo/core/model_factory.py
+++ b/larq_zoo/core/model_factory.py
@@ -41,17 +41,12 @@ class ModelFactory:
 
     input_shape: Optional[Tuple[DimType, DimType, DimType]] = Field(None)
     input_tensor: Optional[tf.Tensor] = Field(None)
-    min_input_resolution: Optional[int] = Field(64)
 
     @property
     def image_input(self) -> tf.Tensor:
         if not hasattr(self, "_image_input"):
             input_shape = utils.validate_input(
-                self.input_shape,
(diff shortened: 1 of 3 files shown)


AdamHillier:
If this is to be added as a field to all the models, I'm not sure that `min_size` is the right name.

AdamHillier:
Could you also try actually building the model (off this branch) with the input shape (32, 32, 3) and see if it successfully builds?

leonoverweel:
Yeah it might be good to call it `min_input_resolution`

MariaHeuss:
I also changed the default in Resnet. Building the model works.

AdamHillier:
If all models can be built with a 32, 32, 3 input shape I'm of half a mind to suggest that we just hard-code the default as 32 rather than creating another field that influences input_shape (which I think is confusing).

AdamHillier:
Also note that I think you can just pass `input_tensor=tf.keras.Input(32, 32, 3)` into the functional API (or alternatively into `ComponentField(...Factory, input_tensor=tf.keras.Input(32, 32, 3))` and it should work?

leonoverweel:
> If all models can be built with a 32, 32, 3 input shape I'm of half a mind to suggest that we just hard-code the default as 32

This would also be my preferred solution if we can confirm that all models can be built this small.

> Also note that I think you can just pass input_tensor=tf.keras.Input(32, 32, 3) into the functional API (or alternatively into ComponentField(...Factory, input_tensor=tf.keras.Input(32, 32, 3)) and it should work?

I don't think this is very intuitive; the solution in this PR (or hardcoding to 32) fixes the problem for users without them having to figure this out.

MariaHeuss:
> If all models can be built with a 32, 32, 3 input shape I'm of half a mind to suggest that we just hard-code the default as 32 rather than creating another field that influences input_shape (which I think is confusing).

Most of the other models can not build with (32, 32, 3) input size. Alexnet can not f.ex.

MariaHeuss:
> Also note that I think you can just pass `input_tensor=tf.keras.Input(32, 32, 3)` into the functional API (or alternatively into `ComponentField(...Factory, input_tensor=tf.keras.Input(32, 32, 3))` and it should work?

I get an Error when I try to do this. I also agree that the approach that I picked is probably more intuitive.

MariaHeuss:
Maybe it would be sensible to set the default value for the ``` min_input_resolution ``` to something smaller. I guess the ModelFactory can be used for all kinds of models and the user should not need to figure out that he has to change this value if he wants to train some small dataset.

leonoverweel:
We could also set it empty and then force each model to override it.

AdamHillier:
How do you feel about just removing that min input shape checking entirely. If a user puts in a small input shape then model building will fail, but I think that's fine, right? We don't actually need to catch this error it seems to me.

PR #137, opened 2020-03-19
Implement QuickNet-XL

koenhelwegen:
Implementation for QuickNet-XL, based on a ResNet-34 with squeeze & excite modules.

The new benchmark results are described in https://github.com/larq/compute-engine/pull/294.

The corresponding model files are released in https://github.com/larq/zoo/releases/tag/quicknet_xl-v0.1.0.


larq_zoo/sota/quicknet_xl.py
```diff
@@ -0,0 +1,179 @@
...
+    )(out)
+    outmult = 2 if strides == 2 else 1
+    out = tf.keras.layers.Dense(
+        C * outmult,
+        activation="sigmoid",
+        kernel_initializer="he_normal",
+        use_bias=False,
+        kernel_regularizer=tf.keras.regularizers.l2(1e-5),
+    )(out)
+    return tf.reshape(out, [-1, 1, 1, C * outmult])
```
lgeiger: Can we reuse the [implementation from `QuickNetLarge`](https://github.com/larq/zoo/blob/master/larq_zoo/sota/quicknet_large.py#L12-L32) here instead?

larq_zoo/sota/quicknet_xl.py
```diff
@@ -0,0 +1,179 @@
...
+                kernel_initializer="glorot_normal",
+            )(residual)
+            residual = tf.keras.layers.BatchNormalization(momentum=0.9, epsilon=1e-5)(
+                residual
+            )
+        else:
+            residual = x
+
+        use_se = filters not in [64, 128]
+        if use_se:
```
lgeiger:
```suggestion
        use_squeeze_and_excite = filters not in (64, 128)
        if use_squeeze_and_excite:
```

larq_zoo/sota/quicknet_xl.py
```diff
@@ -0,0 +1,179 @@
...
+            kernel_quantizer=self.kernel_quantizer,
+            kernel_constraint=self.kernel_constraint,
+            kernel_initializer="glorot_normal",
+            use_bias=False,
+            activation="relu",
+        )(x)
+
+        x = tf.keras.layers.BatchNormalization(momentum=0.9, epsilon=1e-5)(x)
+
+        if use_se:
```
lgeiger:
```suggestion
        if use_squeeze_and_excite:
```

larq_zoo/sota/quicknet_xl.py
```diff
@@ -0,0 +1,179 @@
...
+            x = tf.multiply(x, y)
```
lgeiger:
```suggestion
            x *= y
```

larq_zoo/sota/quicknet_xl.py
```diff
@@ -0,0 +1,179 @@
...
+        activation="sigmoid",
+        kernel_initializer="he_normal",
+        use_bias=False,
+        kernel_regularizer=tf.keras.regularizers.l2(1e-5),
+    )(out)
+    return tf.reshape(out, [-1, 1, 1, C * outmult])
+
+
+@factory
+class QuickNetXLFactory(ModelFactory):
```
lgeiger: Should we subclass `QuickNetLargeFactory`, so we can reuse things like `residual_block_SE`?

larq_zoo/sota/quicknet_xl.py
```diff
@@ -0,0 +1,179 @@
...
+
+    @Field
+    def spec(self) -> Tuple[Sequence[int], Sequence[int]]:
+        spec = {34: ([6, 8, 12, 6], [64, 128, 256, 512])}
+        try:
+            return spec[self.num_layers]
+        except Exception:
+            raise ValueError(f"Only specs for layers {list(self.spec.keys())} defined.")
+
+    def residual_block_SE(
```
lgeiger: I think `residual_block_squeeze_and_excite` would be a better name.

larq_zoo/sota/quicknet_xl.py
jamescook106: As per my comment, it should be a subclass as the only thing that is new is 34 layers instead of 18

koenhelwegen:
> This network is a 34 layer version of QuickNet-Large, so can be implemented via adding a 34 layer spec and the appropriate subclass

Please have a look at https://github.com/larq/zoo/pull/138 first, after that refactor I will rework this.

jamescook106:
> > This network is a 34 layer version of QuickNet-Large, so can be implemented via adding a 34 layer spec and the appropriate subclass
> 
> Please have a look at #138 first, after that refactor I will rework this.

Just reviewed it :)

PR #138, opened 2020-03-22
Refactor quicknets and remove duplication

koenhelwegen:
This PR refactors the implementation of QuickNet and QuickNetLarge so duplication between the two is eliminated. Additionally, the behaviour of the various block functions is made consistent and implicit assumptions regarding things like filter doubling are removed.

New implementation has been tested with pretrained weights and achieves the same accuracy as before.


larq_zoo/sota/quicknet.py
```diff
@@ -104,6 +161,21 @@ def build(self) -> tf.keras.models.Model:
...
 
+        return model
+
+
+@factory
+class QuickNetFactory(QuickNetBaseFactory):
+    """Quicknet - A model designed for fast inference using [Larq Compute Engine](https://github.com/larq/compute-engine)"""
+
+    spec = Field(
+        lambda _: ([2, 3, 4, 4], [64, 128, 256, 512], [False, False, False, False])
```
koenhelwegen: Not sure why the `lambda` is necessary here. @AdamHillier, is this the intended usage or do you have a suggestion?

AdamHillier: It's necessary because anything that's immutable has to be wrapped in a lambda, or *bad things can happen*™. (Bad things in this case means subtle bugs arising from mutable data being modified in one instance of the component class with unpredictable results in other instances. If we require a lambda which returns the data, we can safely generate a new value of the data in each class instance.)

This is a tuple of lists, and lists are mutable, hence the lambda is required. If it were a tuple of tuples of immutable values it would technically be fine, but Zookeeper only checks one level of recursion down and otherwise conservatively assumes mutability, requiring a lambda. The gist is that Zookeeper knows that `None`, `bool`, `int`, `float`, and `str` are all immutable. It also knows that `tuple`s and `frozenset`s of these fundamental types are immutable, and assumes that everything else is mutable.

The lambda can accept 0 or 1 argument (`self`) by the way, so there's no need for the underscore here.

lgeiger: Yes this is intended behaviour.
`Field` doesn't allow mutable default values since this could lead to unexpected behaviour. This is inline with what Python dataclasses do. Take a look at https://docs.python.org/3/library/dataclasses.html#mutable-default-values for an example why mutable default values would be a bad idea.

You could get rid of the lambda here if you use a [`frozenset`](https://docs.python.org/3/library/stdtypes.html#frozenset)  or a tuple of `frozenset`s. Though the lambda is also OK.

koenhelwegen: Okay thanks! I had tried a tuple of tuples, this explains why that throws as well.

larq_zoo/sota/quicknet.py
```diff
@@ -30,42 +31,49 @@ def LCEFirstLayer(filters: int, x: tf.Tensor) -> tf.Tensor:
     return tf.keras.layers.BatchNormalization(momentum=0.9, epsilon=1e-5)(x)
 
 
+def squeeze_and_excite(inp: tf.Tensor, filters: int, r: int = 16):
+    """Squeeze and Excite as per [Squeeze-and-Excitation Networks](https://arxiv.org/abs/1709.01507)"""
```
jamescook106: We should add a reference here (or later) to https://openreview.net/forum?id=BJg4NgBKvH

larq_zoo/sota/quicknet.py
```diff
@@ -30,42 +31,49 @@ def LCEFirstLayer(filters: int, x: tf.Tensor) -> tf.Tensor:
...
+    return tf.reshape(out, [-1, 1, 1, filters])
+
+
 @factory
-class QuickNetFactory(ModelFactory):
-    """Quicknet - A model designed for fast inference using [Larq Compute Engine](https://github.com/larq/compute-engine)"""
+class QuickNetBaseFactory(ModelFactory):
 
-    num_layers: int = Field(15)
+    spec: Tuple[Sequence[int], Sequence[int], Sequence[bool]] = Field(None)
```
jamescook106: The last spec for the SE, although useful here, means that as we transition to being more precise with where we place SE, that this will need to be modified

larq_zoo/sota/quicknet.py
```diff
@@ -30,42 +31,49 @@ def LCEFirstLayer(filters: int, x: tf.Tensor) -> tf.Tensor:
...
+    transition_block: MethodType = Field(None)
     initial_filters: int = Field(64)
```
jamescook106: This needs to be renamed, as technically our initial_filters are 8 and they get expanded 8 times via DW

larq_zoo/sota/quicknet.py
```diff
@@ -30,42 +31,49 @@ def LCEFirstLayer(filters: int, x: tf.Tensor) -> tf.Tensor:
...
+
+
 @factory
-class QuickNetFactory(ModelFactory):
-    """Quicknet - A model designed for fast inference using [Larq Compute Engine](https://github.com/larq/compute-engine)"""
+class QuickNetBaseFactory(ModelFactory):
 
-    num_layers: int = Field(15)
+    spec: Tuple[Sequence[int], Sequence[int], Sequence[bool]] = Field(None)
+    transition_block: MethodType = Field(None)
```
jamescook106: I'm not sure what this is?

larq_zoo/sota/quicknet.py
```diff
@@ -78,20 +86,70 @@ def residual_fast_block(
...
+        residual = tf.keras.layers.MaxPool2D(pool_size=strides, strides=strides)(x)
+        residual = tf.keras.layers.BatchNormalization(momentum=0.9, epsilon=1e-5)(
+            residual
+        )
+        x = self.conv_block(x, infilters, use_squeeze_and_excite, strides)
+        x = tf.keras.layers.add([x, residual])
+
+        return tf.keras.layers.concatenate([residual, x])
+
+    def pointwise_transition_block(
```
jamescook106: Can you please rename this block (or the description). The point of it is more akin to the improvement block in https://arxiv.org/abs/2001.05936

koenhelwegen: The `pointwise` block is just an old-fashioned dimension-changing BiReal-Net block, I think you're talking about the `concat_transition_block`?

larq_zoo/sota/quicknet.py
koenhelwegen: This should point to whichever method is used as transition block (currently, either `concat_transition_block` or `pointwise_transition_block`).

larq_zoo/sota/quicknet.py
koenhelwegen: Yes that's a good point, but I'm not sure if we can/should anticipate this now. Do you have a concrete proposal of how to implement this? Otherwise I think it's fine to modify this again when the time comes.

larq_zoo/sota/quicknet.py
jamescook106: I understand where the name comes from, however I think the renaming of it to something more instructive would be helpful - maybe ```fp_pointwise_transition_block``` or ```fp_improvement_transition_block```

larq_zoo/sota/quicknet.py
```diff
@@ -9,7 +9,8 @@
 from larq_zoo.core.model_factory import ModelFactory
 
 
-def LCEFirstLayer(filters: int, x: tf.Tensor) -> tf.Tensor:
+def stem_module(filters: int, x: tf.Tensor) -> tf.Tensor:
```
jamescook106: The stem module name is already taken in https://arxiv.org/pdf/1908.05858.pdf - please rename

larq_zoo/sota/quicknet.py
jamescook106: I think the transition_block name is unintuitive

larq_zoo/sota/quicknet.py
```diff
@@ -105,6 +110,8 @@ def concat_transition_block(
         """Concat transition block.
 
         Doubles number of filters by concatenating shortcut with x + shortcut.
+        This module is loosely inspired by
+        [MeliusNet](https://arxiv.org/abs/2001.05936).
```
jamescook106: This block was not inspired by this, please remove the reference.

larq_zoo/sota/quicknet.py
koenhelwegen: How about `expansion_module`?

larq_zoo/sota/quicknet.py
koenhelwegen: Why? Do you have any alternative in mind?

larq_zoo/sota/quicknet.py
lgeiger: I would prefer `stem_module` as well. STEM is a pretty general name for the first few input layer also used in https://arxiv.org/pdf/1812.01187.pdf and I think since it is defined in the `quicknet` file it is clear that it relates to quicknet and not to an other model.

larq_zoo/sota/quicknet.py
```diff
@@ -30,42 +32,53 @@ def LCEFirstLayer(filters: int, x: tf.Tensor) -> tf.Tensor:
...
+    """Squeeze and Excite as per [Squeeze-and-Excitation Networks](https://arxiv.org/abs/1709.01507).
+
+    Use of S&E in BNNs was pioneered in [Training binary neural networks with
+    real-to-binary convolutions](https://openreview.net/forum?id=BJg4NgBKvH).
+    """
+    C = inp.get_shape().as_list()[-1]
+
+    out = utils.global_pool(inp)
+    out = tf.keras.layers.Dense(
+        C // r,
```
lgeiger:
```suggestion
    out = utils.global_pool(inp)
    out = tf.keras.layers.Dense(
        inp.shape[-1] // r,
```

larq_zoo/sota/quicknet.py
lgeiger: > I think the transition_block name is unintuitive

Transition block is quite widely used in literature, do you have an alternative?

larq_zoo/sota/quicknet.py
```diff
@@ -78,20 +91,72 @@ def residual_fast_block(
...
-            return tf.keras.layers.add([x, residual])
+
+        if use_squeeze_and_excite:
+            x *= y
+
+        return x
+
+    def residual_block(self, x: tf.Tensor, use_squeeze_and_excite: bool) -> tf.Tensor:
+        """Standard residual block, without strides or filter changes."""
+        infilters = x.get_shape().as_list()[-1]
```
lgeiger:
```suggestion
        infilters = x.shape[-1]
```

larq_zoo/sota/quicknet.py
```diff
@@ -78,20 +91,72 @@ def residual_fast_block(
...
+    def concat_transition_block(
+        self, x: tf.Tensor, filters: int, strides: int, use_squeeze_and_excite: bool
+    ) -> tf.Tensor:
+        """Concat transition block.
+
+        Doubles number of filters by concatenating shortcut with x + shortcut.
+        This module is loosely inspired by
+        [MeliusNet](https://arxiv.org/abs/2001.05936).
+        """
+        infilters = x.get_shape().as_list()[-1]
```
lgeiger:
```suggestion
        infilters = x.shape[-1]
```

larq_zoo/sota/quicknet.py
```diff
@@ -78,20 +91,72 @@ def residual_fast_block(
...
+        x = stem_module(self.stem_filters, self.image_input)
 
-        for block, (layers, filters) in enumerate(zip(*self.spec)):
+        for block, (layers, filters, use_squeeze_and_excite) in enumerate(
+            zip(*self.spec)
+        ):
             for layer in range(layers):
-                strides = 1 if block == 0 or layer != 0 else 2
-                x = self.residual_fast_block(x, filters, strides=strides)
+                if filters == x.get_shape().as_list()[-1]:
```
lgeiger:
```suggestion
                if filters == x.shape[-1]:
```

larq_zoo/sota/quicknet.py
```diff
@@ -104,6 +169,21 @@ def build(self) -> tf.keras.models.Model:
...
 
+        return model
+
+
+@factory
+class QuickNetFactory(QuickNetBaseFactory):
+    """Quicknet - A model designed for fast inference using [Larq Compute Engine](https://github.com/larq/compute-engine)"""
+
+    spec = Field(
+        lambda: ([2, 3, 4, 4], [64, 128, 256, 512], [False, False, False, False])
```
lgeiger: I am not sure if we should really have this as a field, or if we should just make the total number of layers configurable. I don't have a strong opinion about this, but for me it is not really intuitive what the different config values are here.

koenhelwegen: Yes I'm also not too happy about this. We could split it into `blocks_per_section`, `section_filters`, `use_attention_in_section` or something like that?

larq_zoo/sota/quicknet.py
leonoverweel: `ScienceTechnologyEngineeringMath_module` maybe?

PR #140, opened 2020-03-23
Add names to quicknet models

koenhelwegen:


AdamHillier:
This doesn't super matter but I don't think this needs to be a field as opposed to simply a class attribute, because I don't think anyone will ever override the value.

lgeiger:
> This doesn't super matter but I don't think this needs to be a field as opposed to simply a class attribute, because I don't think anyone will ever override the value.

I agree. This doesn't need to be a field.

leonoverweel:
I can see a scenario where you're using a QuickNet as an encoder for a task and you want to rename it something generic like `encoder` to make different encoders have matching naming schemes for later layer accessing. Can't hurt to make it a `Field` right?

koenhelwegen:
> I can see a scenario where you're using a QuickNet as an encoder for a task and you want to rename it something generic like `encoder` to make different encoders have matching naming schemes for later layer accessing. Can't hurt to make it a `Field` right

I agree, but we're not doing it for any of the other models atm. I've removed it now; if we want to allow users to modify this, I propose to make a separate PR for all of zoo.

PR #144, opened 2020-04-09
Real-to-binary models

timdebruin:
Note: training procedure will be a separate PR

TODO:
 - [x] weight files
  ~~- strong baseline BNN~~
   - r2b BNN
 - [x] training json for altair 
   ~~- strong baseline BNN~~
   - r2b BNN


larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,494 @@
...
+
+        def get_config(self):
+            return {"output_dim": self.output_dim}
+
+    def _scale_binary_conv_output(
+        self, conv_input: tf.Tensor, conv_output: tf.Tensor, name: str
+    ):
+        """"The way in which the output of the binary convolution is scaled is the only structural difference
+        between the StrongBaseline networks and the Real-to-Binary networks. This function accepts all inputs
+        required for either function."""
```
leonoverweel, line 182:
```suggestion
        """Flexible wrapper for the `LearnedRescaleLayer`.
        
        The way in which the output of the binary convolution is scaled is the only
        structural difference between the StrongBaseline networks and the Real-to-Binary
        networks. This function accepts all inputs required for either function.
        """
```

(Edited for max-width of 88 chars.)

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,494 @@
...
+
+        # Batch Normalization
+        conv_input = tf.keras.layers.BatchNormalization(
+            momentum=self.momentum, name=f"{name}_batch_norm"
+        )(x)
+
+        # Convolution
+        conv_output = lq.layers.QuantConv2D(
+            out_channels,
+            3,
```
leonoverweel, line 212:
```suggestion
            kernel_size=3,
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,494 @@
...
+        def call(self, inputs, **kwargs):
+            return inputs * (self.scale_h * self.scale_w * self.scale_c)
+
+        def compute_output_shape(self, **kwargs):
+            return self.output_dim
+
+        def get_config(self):
+            return {"output_dim": self.output_dim}
+
+    def _scale_binary_conv_output(
```
leonoverweel, line 177: (Unlike the block methods, I'd be happy to keep this wrapper as a private method.)

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,494 @@
...
+        # Skip connection
+        return tf.keras.layers.Add(name=f"{name}_skip_add")([x, shortcut_add])
+
+    def _block(self, x, downsample=False, name=""):
+        """Full binary block from Figure 1 (Left) of Matrinez et al. (2019)."""
+
+        x = self._half_binary_block(x, downsample=downsample, name=f"{name}a")
+        x = self._half_binary_block(x, downsample=False, name=f"{name}b")
+
+        # Add explicit name to the block output for attention matching (Section 4.2 of Martinez et al.)
```
leonoverweel, line 237:
```suggestion
        # Add explicit name to the block output for attention matching (Section 4.2 of 
        # Martinez et al.)
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,494 @@
...
+                file="r2b_strong_baseline_bnn_weights.h5",
+                file_hash="",  # TODO
+            )
+        else:
+            weights_path = utils.download_pretrained_model(
+                model="baseline_bnn",
+                version="v0.1.0",
+                file="r2b_strong_baseline_bnn_weights_notop.h5",
+                file_hash="",  # TODO
+            )
```
leonoverweel, line 261: Is the plan to keep these as to do's until we add training code?

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,494 @@
...
+        return weights_path
+
+
+@factory
+class RealToBinNetFactory(StrongBaselineNetFactory):
+    def _scale_binary_conv_output(
+        self, conv_input: tf.Tensor, conv_output: tf.Tensor, name: str
+    ):
+        """Data-dependent (squeeze and excite style) scaling of the convolution output as described in
+        Section 4.3 of Martinez at. al.
```
leonoverweel, line 271:
```suggestion
        """Squeeze-and-Excite-style scaling.
        
        Data-dependent scaling of the convolution output, as described in Section 4.3
        of Martinez at. al.
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,494 @@
...
+
+@factory
+class ResNet18Factory(_SharedBaseFactory):
+    """Constructor for a ResNet18 with layer names matching Real-to-Bin nets."""
+
+    def _block(self, x, downsample=False, name=""):
+        """One full residual block, consisting of two convolutions.
+
+        This follows the definition of a "block" from Figure 1 (Left) of Martinez et al.
+        (2019).
```
leonoverweel, line 332:
```suggestion
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,494 @@
...
+                padding="same",
+                name=f"{name}{convolution}_conv2d",
+            )(x)
+            x = tf.keras.layers.BatchNormalization(
+                momentum=self.momentum, name=f"{name}{convolution}_batch_norm"
+            )(x)
+            x = tf.keras.layers.Activation("relu", name=f"{name}{convolution}_relu")(x)
+
+        x = tf.keras.layers.Add(name=f"{name}_skip_add")([x, shortcut_add])
+        return tf.keras.layers.Activation("relu", name=f"{name}_out")(x)
```
leonoverweel, line 355:
```suggestion
        return tf.keras.layers.ReLU(name=f"{name}_out")(x)
```

larq_zoo/literature/real_to_bin_nets.py
timdebruin: No I think we should add them before that (see also top comment of the PR)

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,494 @@
...
+
+class _SharedBaseFactory(ModelFactory):
+    """Base configuration and blocks shared across ResNet, StrongBaselineNets and Real-to-Bin Nets."""
+
+    model_name: str = Field()
+    momentum: float = Field(0.99)
+    kernel_initializer: str = Field("glorot_normal")
+    kernel_regularizer: Optional[tf.keras.regularizers.Regularizer] = Field(None)
+
+    def _first_block(self, x, use_prelu=True, name=""):
```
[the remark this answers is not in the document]
timdebruin: Fixed in eb6b2dd

larq_zoo/literature/real_to_bin_nets.py
timdebruin: I changed this to:

```
        """Data-dependent convolution scaling.

        Scales the output of the convolution in the (squeeze-and-excite
        style) data-dependent way described in Section 4.3 of Martinez at. al.
        """
```
In the paper the link to squeeze and excite is not made very explicitly. They introduce it as their novel 'data-dependent scaling trick'. So if someone is looking for this magical data dependent scaling trick I think that should be the headline.

larq_zoo/literature/real_to_bin_nets.py
leonoverweel: Ahh makes sense; wasn't sure if the to do's were for this PR or combined for all PRs for this model.

larq_zoo/literature/real_to_bin_nets.py
leonoverweel: Nice, makes sense.

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+        )
+
+        # Load weights.
+        if self.weights == "imagenet":
+            model.load_weights(self._get_weights_path())
+        elif self.weights is not None:
+            model.load_weights(self.weights)
+        return model
+
+    def _get_weights_path(self):
```
AdamHillier, line 120:
```suggestion
    @abc.abstractmethod
    def _get_weights_path(self) -> str:
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+        """Implements the learned activation rescaling XNOR-Net++ style.
+
+        This is used to scale the outputs of the binary convolutions in the Strong
+        Baseline networks. [(Bulat & Tzimiropoulos,
+        2019)](https://arxiv.org/abs/1909.13863)
+        """
+
+        def __init__(
+            self,
+            output_dim: tuple,
```
AdamHillier, line 143:
```suggestion
            output_dim: Tuple[int, int, int],
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+            super().build(input_shapes)
+
+        def call(self, inputs, **kwargs):
+            return inputs * (self.scale_h * self.scale_w * self.scale_c)
+
+        def compute_output_shape(self, **kwargs):
+            return self.output_dim
+
+        def get_config(self):
+            return {"output_dim": self.output_dim}
```
AdamHillier, line 186: Do we need to serialise the regulariser here too?

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+
+    def _scale_binary_conv_output(
+        self, conv_input: tf.Tensor, conv_output: tf.Tensor, name: str
+    ):
```
AdamHillier, line 190:
```suggestion
    ) -> tf.Tensor:
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+        The Strong Baseline network uses the learned static rescale layer of
+        Bulat & Tzimiropoulos
+        """
+        return self.LearnedRescaleLayer(
+            conv_output.shape[1:],
+            regularizer=self.kernel_regularizer,
+            name=f"{name}_rescale",
+        )(conv_output)
+
+    def half_binary_block(self, x: tf.Tensor, downsample: bool = False, name: str = ""):
```
AdamHillier, line 206:
```suggestion
    def half_binary_block(
        self, x: tf.Tensor, downsample: bool = False, name: str = ""
    ) -> tf.Tensor:
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+        x = tf.keras.layers.PReLU(shared_axes=[1, 2], name=f"{name}_prelu")(x)
+
+        # Skip connection
+        return tf.keras.layers.Add(name=f"{name}_skip_add")([x, shortcut_add])
+
+    def block(self, x, downsample=False, name=""):
+        """Full binary block from Figure 1 (Left) of Matrinez et al.
+
+        (2019).
+        """
```
AdamHillier, line 256:
```suggestion
        """Full binary block from Figure 1 (Left) of Matrinez et al. (2019)."""
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+        # Martinez et al.)
+        return tf.keras.layers.Lambda(lambda x: x, name=f"{name}_out")(x)
+
+    def _get_weights_path(self):
+        if (
+            not self.kernel_quantizer == "ste_sign"
+            and self.input_quantizer == "ste_sign"
+        ):
+            raise NotImplementedError(
+                f"{self.model_name} only has stored weights for the BNN variant"
```
AdamHillier, line 271:
```suggestion
                f"{self.model_name} only has ImageNet weights for the BNN variant"
```

larq_zoo/literature/real_to_bin_nets.py
AdamHillier: Also should we maybe make it clear that these are imagenet weights and call it `_get_imagenet_weights_path()`?

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+        # binary convolution rescaling
+        x = self._scale_binary_conv_output(conv_input, conv_output, name)
+
+        # PReLU activation
+        x = tf.keras.layers.PReLU(shared_axes=[1, 2], name=f"{name}_prelu")(x)
+
+        # Skip connection
+        return tf.keras.layers.Add(name=f"{name}_skip_add")([x, shortcut_add])
+
+    def block(self, x, downsample=False, name=""):
```
AdamHillier, line 252:
```suggestion
    def block(self, x: tf.Tensor, downsample: bool = False, name: str = "") -> tf.Tensor:
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+                file_hash="",  # TODO
+            )
+        return weights_path
+
+
+@factory
+class RealToBinNetFactory(StrongBaselineNetFactory):
+    def _scale_binary_conv_output(
+        self, conv_input: tf.Tensor, conv_output: tf.Tensor, name: str
+    ):
```
AdamHillier, line 294:
```suggestion
    ) -> tf.Tensor:
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+            [conv_output, scales]
+        )
+
+    def _get_weights_path(self):
+        if (
+            not self.kernel_quantizer == "ste_sign"
+            and self.input_quantizer == "ste_sign"
+        ):
+            raise NotImplementedError(
+                f"{self.model_name} only has stored weights for the BNN variant"
```
AdamHillier, line 334:
```suggestion
                f"{self.model_name} only has ImageNet weights for the BNN variant"
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+                file_hash="4ec47abf1a4da5c65f4908076257e8d5c812673891089a88c9d9e84e949d1dab",
+            )
+        return weights_path
+
+
+@factory
+class ResNet18Factory(_SharedBaseFactory):
+    """Constructor for a ResNet18 with layer names matching Real-to-Bin nets."""
+
+    def block(self, x, downsample=False, name=""):
```
AdamHillier, line 357:
```suggestion
    def block(self, x: tf.Tensor, downsample: bool = False, name: str = "") -> tf.Tensor:
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+                momentum=self.momentum, name=f"{name}{convolution}_batch_norm"
+            )(x)
+            x = tf.keras.layers.Activation("relu", name=f"{name}{convolution}_relu")(x)
+
+        x = tf.keras.layers.Add(name=f"{name}_skip_add")([x, shortcut_add])
+        return tf.keras.layers.ReLU(name=f"{name}_out")(x)
+
+
+@factory
+class StrongBaselineNetBAN(StrongBaselineNetFactory):
```
AdamHillier, line 388: Do you mind if we stick to appending `Factory` to the names of these (including the five or so directly below)? I know it makes the names quite long but I like the consistency.

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+        ImageNet), or the path to the weights file to be loaded.
+    include_top: whether to include the fully-connected layer at the top of the network.
+    num_classes: optional number of classes to classify images into, only to be
+        specified if `include_top` is True, and if no `weights` argument is specified.
+
+    # Returns
+    A Keras model instance.
+
+    # Raises
+    ValueError: in case of invalid argument for `weights`, or invalid input shape.
```
AdamHillier, line 479: Is it possible for a `NotImplementedError` to be raised?

larq_zoo/literature/real_to_bin_nets.py
timdebruin: We actually only provide weights for the BNN version of the R2B net. So this method (which says there are no weights) is the default. It is only overwritten for the R2B subclass and then only provides Imagenet weights.

AdamHillier: Ah I see, makes sense (for not being an ABC). How do you feel about renaming the function?

AdamHillier: And perhaps this should be a `ValueError`, because although it's true that the method is not implemented, a user would only see this if they put in the wrong value for the `weights` argument.

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,490 @@
...
+        return tf.keras.layers.MaxPool2D(
+            3, strides=2, padding="same", name=f"{name}_pool"
+        )(x)
+
+    def last_block(self, x: tf.Tensor, name: str = "") -> tf.Tensor:
+        """Last block, shared across ResNet, StrongBaselineNet and Real-to-Bin nets."""
+
+        x = tf.keras.layers.GlobalAvgPool2D(name=f"{name}_global_pool")(x)
+        x = tf.keras.layers.Dense(
+            self.num_classes, activation=None, name=f"{name}_logits",
```
lgeiger, line 58:
```suggestion
            self.num_classes, name=f"{name}_logits",
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,490 @@
...
+            assert (
+                len(output_dim) == 3
+            ), f"expected 3-dimensional output size, got {len(output_dim)}-dimensional"
+            self.output_dim = output_dim
+            self.kernel_regularizer = regularizer
+
+        def build(self, input_shapes):
+            self.scale_h = self.add_weight(
+                name="scale_h",
+                shape=(self.output_dim[0], 1, 1),
```
lgeiger, line 157: Why do we need to set `self.output_dim`? Can't we just get this information from `input_shapes`?

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,490 @@
...
+                trainable=True,
+            )
+
+            super().build(input_shapes)
+
+        def call(self, inputs, **kwargs):
+            return inputs * (self.scale_h * self.scale_w * self.scale_c)
+
+        def compute_output_shape(self, **kwargs):
+            return self.output_dim
```
lgeiger, line 183: Why not:
```suggestion
        def compute_output_shape(self, input_shape):
            return input_shape
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,490 @@
...
+@factory
+class StrongBaselineNetFactory(_SharedBaseFactory):
+    """Constructor for the strong baseline network (Section 4.1 of Martinez et al.)."""
+
+    scaling_r: int = 8
+
+    input_quantizer: QuantizerType = Field(None)
+    kernel_quantizer: QuantizerType = Field(None)
+
+    class LearnedRescaleLayer(tf.keras.layers.Layer):
```
lgeiger, line 133: I think I would define this layer outside of the class, but this is just personal preference.

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,491 @@
...
+    def last_block(self, x, name=""):
+        """Last block, shared across ResNet, StrongBaselineNet and Real-to-Bin nets."""
+
+        x = tf.keras.layers.GlobalAvgPool2D(name=f"{name}_global_pool")(x)
+        x = tf.keras.layers.Dense(
+            self.num_classes, activation=None, name=f"{name}_logits",
+        )(x)
+        return tf.keras.layers.Softmax(name=f"{name}_probs", dtype=tf.float32)(x)
+
+    def block(self, x, downsample=False, name=""):
```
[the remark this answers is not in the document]
timdebruin: Implemented in d4b4c78

larq_zoo/literature/real_to_bin_nets.py
timdebruin: Implemented in d4b4c78

larq_zoo/literature/real_to_bin_nets.py
timdebruin: It was, fixed in d4b4c78

larq_zoo/literature/real_to_bin_nets.py
timdebruin: You're right, fixed in d4b4c78

larq_zoo/literature/real_to_bin_nets.py
timdebruin: Fixed in d4b4c78

larq_zoo/literature/real_to_bin_nets.py
timdebruin: I can see your point (in fact, that is how it was earlier). The reasoning for having it here is for when people have read the paper and then look at our implementation:
R2B net is architecturally just a StrongBaselineNet which has a different conv output scaling strategy. (And StrongBaselineNet is just a ResnetE18 with some modifications). The current implementation is I think not the best software design but it is the most pedagogical way of defining things to understand the paper (and if you understand the paper to understand the code).

larq_zoo/literature/real_to_bin_nets.py
timdebruin: It is now a `ValueEror` and the name is changed (I forgot that there was an explicit check that the name is `imagenet` before calling this function).

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,472 @@
...
+        This is used to scale the outputs of the binary convolutions in the Strong
+        Baseline networks. [(Bulat & Tzimiropoulos,
+        2019)](https://arxiv.org/abs/1909.13863)
+        """
+
+        def __init__(
+            self, regularizer: Optional[tf.keras.regularizers.Regularizer], **kwargs,
+        ) -> None:
+            super().__init__(**kwargs)
+            self.kernel_regularizer = regularizer
```
lgeiger, line 144: If you do the following, you don't need to overwrite `from_config`.
```suggestion
            self.kernel_regularizer = tf.keras.regularizers.get(regularizer)
```

larq_zoo/literature/real_to_bin_nets.py
```diff
@@ -0,0 +1,470 @@
...
+                "output_dim": self.output_dim,
+                "regularizer": self.kernel_regularizer.get_config(),
+            }
+
+        @classmethod
+        def from_config(cls, config):
+            return cls(
+                output_dim=config["output_dim"],
+                regularizer=tf.keras.regularizers.deserialize(config["regularizer"]),
+            )
```
lgeiger, line 188: See above:
```suggestion
        def get_config(self):
            return {
                **super().get_config(),
                "regularizer": tf.keras.regularizers.serialize(self.kernel_regularizer),
            }
```

timdebruin:
> Nice, I think this should be ready to merge very soon! Can you just [add a test case](https://github.com/larq/zoo/blob/master/tests/models_test.py#L30) for these models so we can make sure they keep building correctly over time?

c28db15 adds the test for the R2B BNN. The other models do not have direct (user facing) build functions but are there to demonstrate the training process. I will add tests for those models in the training PR (#150).

PR #147, opened 2020-04-11
Simplify testing and dummy data generation

lgeiger:
This uses `tfds.testing.mock_data` to simplify our training integration tests.


tests/train_test.py
```diff
@@ -1,23 +1,29 @@
...
+from zookeeper.tf.dataset import TFDSDataset
 
 from larq_zoo.training import basic_experiments
-from tests import dummy_datasets  # noqa
 
 
 @pytest.mark.parametrize("command", list(basic_experiments.cli.commands.keys()))
-def test_cli(command):
+@tfds.testing.mock_data(num_examples=2, data_dir="gs://tfds-data/dataset_info")
+@mock.patch.object(TFDSDataset, "num_examples", return_value=2)
```
lgeiger, line 13: We patch `TFDSDataset.num_examples` because otherwise training would run for the size of the original dataset.

tests/train_test.py
```diff
@@ -1,23 +1,29 @@
...
 from click.testing import CliRunner
+from zookeeper.tf.dataset import TFDSDataset
 
 from larq_zoo.training import basic_experiments
-from tests import dummy_datasets  # noqa
 
 
 @pytest.mark.parametrize("command", list(basic_experiments.cli.commands.keys()))
-def test_cli(command):
+@tfds.testing.mock_data(num_examples=2, data_dir="gs://tfds-data/dataset_info")
```
lgeiger, line 12: Setting `data_dir` to GCS, allows us to read the dataset info without needing to have the metadata files available locally.

tests/train_test.py
```diff
@@ -1,23 +1,29 @@
...
-            "dataset=DummyOxfordFlowers",
+            "dataset=ImageNet",
             "epochs=1",
             "batch_size=2",
             "validation_frequency=5",
             "--no-use_tensorboard",
             "--no-use_model_checkpointing",
             "--no-save_weights",
         ],
+        catch_exceptions=False,
```
lgeiger, line 27: This will properly log a stack trace when the CLI command fails.

PR #150, opened 2020-04-15
Real to bin training

timdebruin:
This implements the training code for real-to-binary models (both the full model and the strong baseline). For the networks, this PR depends on #144 .

The training code includes:
 - [x] Code to run multi-stage experiments 
 - [x] Code to use teacher-student style knowledge distillation 
 - [x] Code for the strong baseline experiments
   - [x] BAN and BNN experiments
   - [x] experiment sequence
 - [x] Code for the full R2B experiments
   - [x] RN18, FP, BAN, BNN experiments 
   - [x] experiment sequence


timdebruin:
Since #144 is now merged, this is now ready for review.

AdamHillier:
Although I'm on board with removing the 1.14 tests in order to fully test and review this PR, I don't think we should merge this until we've removed 1.14 support everywhere (first in Larq, released v1.0 of Larq, and updated Zoo to require Larq v1.0).

leonoverweel:
I don't think I agree. We don't want to rush out a 1.0 release for Larq just for this—that release should coincide with some bigger new feature(s) + a marketing push in the future—, and I don't think it we should keep this PR blocked for that. This PR in turn blocks the 1.0 release of Zoo, which we want to get out in the next few days, and which would then also be blocked until the eventual Larq 1.0 release.

What's the problem with dropping TF 1.14 support in Zoo before we drop it in Larq?

timdebruin:
It turned out that the test was  breaking for 1.15 aswell. I fixed that and re-enabled the 1.14 test (which I hope is now also fixed) so we can uncouple dropping 1.14 support from this PR.

larq_zoo/training/multi_stage_experiments.py
```diff
@@ -0,0 +1,257 @@
...
+    StrongBaselineNetBANFactory,
+    StrongBaselineNetBNNFactory,
+)
+from larq_zoo.training.datasets import ImageNet
+from larq_zoo.training.knowledge_distillation.multi_stage_training import (
+    LarqZooModelTrainingPhase,
+    MultiStageExperiment,
+)
+
+# --------- Learning rate schedules -------------
```
leonoverweel, line 18: It might be nice to split these (or at least the `CosineDecayWithWarmup`) out into a separate `learning_rates.py` or `utils.py` file under `larq_zoo.training`, since they're useful beyond just R2B.

larq_zoo/training/multi_stage_experiments.py
```diff
@@ -0,0 +1,257 @@
...
+class TrainR2BStrongBaselineBAN(LarqZooModelTrainingPhase):
+    stage = Field(0)
+
+    dataset = ComponentField(ImageNet)
+
+    learning_rate: float = Field(1e-3)
+    learning_rate_decay: float = Field(0.1)
+    epochs: int = Field(75)
+    batch_size: int = Field(8)
+    amount_of_images: int = Field(1281167)
```
leonoverweel, line 109: Do we need this hardcoded here, or can we get it from `self.dataset.num_examples("train")` where it's needed?

larq_zoo/training/multi_stage_experiments.py
```diff
@@ -0,0 +1,257 @@
...
+    warmup_duration: int = Field(5)
+
+    @property
+    def steps_per_epoch(self):
+        return self.amount_of_images // self.batch_size
```
leonoverweel, line 114: I think it might make more sense to add this to the base `zoo.train.TrainLarqZooModel` class.

larq_zoo/training/multi_stage_experiments.py
```diff
@@ -0,0 +1,257 @@
...
+    )
+
+    student_model = ComponentField(ResNet18FPFactory)
+
+
+@task
+class TrainR2BBFP(TrainFPResnet18):
+    stage = Field(1)
+    learning_rate: float = Field(1e-3)
+    learning_rate_decay: float = Field(0.3)  # TODO double check
```
leonoverweel, line 177: `# TODO double check`

larq_zoo/training/multi_stage_experiments.py
timdebruin: Done in be71e12

larq_zoo/training/multi_stage_experiments.py
timdebruin: Good point! Done in be71e12

larq_zoo/training/multi_stage_experiments.py
timdebruin: Also a good point, done in be71e12

larq_zoo/training/multi_stage_experiments.py
timdebruin: Removed in be71e12

PR #153, opened 2020-04-17
Fix workflow syntax

lgeiger:
#152 second try


lgeiger:
> Do the quotation marks really make a difference?

This is how they do it at: https://github.com/pypa/gh-action-pypi-publish#usage

lgeiger:
Looks like this worked: https://github.com/larq/zoo/actions/runs/80821831

Issue #157, opened 2020-04-24
QuickNet no-top models pretrained weights are not working as expected


itayalfia:
### Describe the bug
Tried freezing the no-top quicknet models, and training a linear classifier on top of them, in order to classify images from the Imagenette dataset (10 easy classes from ImageNet).

Because the pretrained zoo models are trained on the superset of this dataset, I expected the pretrained embedders to perform very well, but they did not succeed in reaching above 50% accuracy.

However, when I manually cut the full models, the embedders work as expected and reach 95% easily, hinting the problem is with the no-top pretrained weights.

### To Reproduce
Run the code below with the following configurations:
`QuickNetBugTest cut_full_model=True`
`QuickNetBugTest cut_full_model=False`
`QuickNetLargeBugTest cut_full_model=True`
`QuickNetLargeBugTest cut_full_model=False`
`QuickNetXLBugTest cut_full_model=True`
`QuickNetXLBugTest cut_full_model=False`

(`cut_full_model` param determines if the pretrained no-top model is used, or the pretrained full model is taken and cut before the global pooling. The models are trained for 3 epochs in the example but even if trained more the no-top model does not improve much.)

```python
import tensorflow as tf
from tensorflow import keras
import tensorflow_datasets as tfds
from larq_zoo.training.data import preprocess_image_bytes
from larq_zoo.sota import QuickNet, QuickNetLarge, QuickNetXL
from zookeeper import cli, task, Field
from typing import Callable, Tuple, Optional


class EmbedderWrapperModel(keras.Model):
    def __init__(self, zoo_class: Callable[..., keras.Model],
                 input_shape: int, num_classes: int, dynamic=False,
                 finetune_basenet=True, pretrained_basenet=True, cut_layer_name: Optional[str] = None):
        super(EmbedderWrapperModel, self).__init__(dynamic=dynamic)

        self.basenet = self._get_basenet(zoo_class, input_shape, finetune_basenet, pretrained_basenet, cut_layer_name)
        global_pool_shape = self.basenet.output_shape[1], self.basenet.output_shape[2]

        self.batch_norm = keras.layers.BatchNormalization(momentum=0.9, epsilon=1e-5)
        self.global_pool = keras.layers.AveragePooling2D(pool_size=global_pool_shape)
        self.dense_softmax = keras.layers.Dense(num_classes, activation=tf.nn.softmax)

    def _get_basenet(self, zoo_class: Callable[..., keras.Model], input_shape: int,
[...]

PR #158, opened 2020-04-24
Add test to ensure that notop-weight are correct

lgeiger:
This PR adds a check to ensure that weights of the notop model are correct.

This test currently fails on the following models:
- [x] BinaryAlexNet
- [x] BinaryDenseNet37Dilated
- [x] MeliusNet22
- [x] DoReFaNet
- [x] XNORNet
- [x] QuickNet
- [x] QuickNetLarge
- [x] QuickNetXL


leonoverweel:
Closes #157.

PR #167, opened 2020-04-30
Replace image shape call

leonoverweel:
This should fix the flakey tests which are (probably) failing because of some PNG sample training images pretending to be JPEG.

(Thanks @AdamHillier for the `get_shape_of_image()` snippet.)


leonoverweel:
Ah looks like @AdamHillier pushed this too in #166, closing.

PR #175, opened 2020-06-08
Type hints update

timdebruin:
This makes multi-stage training compatible with newer zookeeper versions

### DIFF
--- a/larq_zoo/training/knowledge_distillation/multi_stage_training.py
+++ b/larq_zoo/training/knowledge_distillation/multi_stage_training.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 from pathlib import Path
-from typing import List, Optional, Union
+from typing import List, Optional, Sequence, Union
 
 import tensorflow as tf
 from zookeeper import ComponentField, Field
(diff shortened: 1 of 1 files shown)


larq_zoo/training/knowledge_distillation/multi_stage_training.py
```diff
@@ -106,7 +106,11 @@ class LarqZooModelTrainingPhase(TrainingPhase):
     output_matching_train_teacher: bool = Field(False)
 
     @Field
-    def loss(self) -> Optional[Union[tf.keras.losses.Loss, str]]:
+    def loss(
+        self,
+    ) -> Optional[
+        Union[Sequence[Union[tf.keras.losses.Loss, str]], tf.keras.losses.Loss, str]
+    ]:
```
lgeiger, line 113: Since the annotation is defined on the superclass, how do you feel about completely removing it here?
```suggestion
    def loss(self):
```

larq_zoo/training/knowledge_distillation/multi_stage_training.py
```diff
@@ -1,6 +1,6 @@
 from datetime import datetime
 from pathlib import Path
-from typing import List, Optional, Union
+from typing import List, Optional, Sequence, Union
```
lgeiger, line 3:
```suggestion
from typing import List, Optional
```

lgeiger:
Thanks for the fix. The [test failure](https://github.com/larq/zoo/pull/174/checks?check_run_id=728960086) of #174 might be caused by this as well.

PR #179, opened 2020-06-15
Cleanup tests

lgeiger:
This PR cleansup some of our tests by relying on pytest fixtures and parametrized tests. This also fixes a bug in the global pooling test where `data_format="channels_last"` was never tested.

### DIFF
--- a/tests/models_test.py
+++ b/tests/models_test.py
@@ -1,4 +1,3 @@
-import functools
 import os
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


tests/models_test.py
```diff
@@ -8,25 +7,21 @@
...
 
 
-def keras_test(func):
-    """Function wrapper to clean up after TensorFlow tests.
-    # Arguments
-        func: test function to clean up after.
-    # Returns
-        A function wrapping the input function.
-    """
+@pytest.fixture(autouse=True)
```
lgeiger, line 10: Autouse makes sure that the fixture is used on every test

tests/models_test.py
```diff
@@ -8,25 +7,21 @@
...
+    request.addfinalizer(keras.backend.clear_session)
 
-    @functools.wraps(func)
-    def wrapper(*args, **kwargs):
-        output = func(*args, **kwargs)
-        keras.backend.clear_session()
-        return output
 
-    return wrapper
+@pytest.fixture(scope="module")
```
lgeiger, line 15: `scope="module"` will cache the result of the fixture for all tests in this file

tests/utils_test.py
```diff
@@ -1,24 +1,17 @@
...
+def test_global_pool(input_res, data_format):
+    shape = (3, *input_res) if data_format == "channels_first" else (*input_res, 3)
+    inp = tf.keras.Input(shape=shape)
+    x = tf.keras.layers.Conv2D(6, 3, 2, data_format=data_format)(inp)
+    x = utils.global_pool(x, data_format=data_format)
+    x = tf.keras.layers.Dense(2)(x)
+    model = tf.keras.Model(inp, x)
 
-    for data_format in ["channels_first", "channels_last"]:
-        fixed_model = build_model()
```
lgeiger, line 19: `data_format` was never passed into the model builder

lgeiger:
> What was it that prompted this, did you come across a bug that the tests were not covering?

I briefly looked into adding proper distribution strategy tests for larq and then thought I came across a way to speed up our tests. This didn't turn out to be true, but during waiting for some zoo tests I ended up refactoring this and noticed the testing bug in the global avg pool test.

PR #248, opened 2020-10-15
Add QuickNet v1 architecture

lgeiger:
This PR adds a new version of QuickNet. I also renamed to models to `QuickNetSmall`, `QuickNet` and `QuickNetLarge` and added the simple training loop used to train the models.

### DIFF
--- a/larq_zoo/sota/quicknet.py
+++ b/larq_zoo/sota/quicknet.py
@@ -246,6 +246,9 @@ def QuickNet(
     ```summary
     sota.QuickNet
     ```
+    ```plot-altair
+    /plots/quicknet.vg.json
+    ```
 
     # ImageNet Metrics
 
@@ -300,6 +303,9 @@ def QuickNetLarge(
     ```summary
     sota.QuickNetLarge
     ```
+    ```plot-altair
+    /plots/quicknet_large.vg.json
+    ```
 
     # ImageNet Metrics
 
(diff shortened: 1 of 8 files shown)


lgeiger:
This is now ready for review. I also manually verified that reloaded QuickNet reaches the correct ImageNet accuracy.

larq_zoo/sota/quicknet.py
```diff
@@ -1,69 +1,56 @@
...
-
-
-class QuickNetBaseFactory(ModelFactory, abc.ABC):
-    name: str = "model"
-    blocks_per_section: Sequence[int] = Field()
-    section_filters: Sequence[int] = Field()
-    use_squeeze_and_excite_in_section: Sequence[bool] = Field()
-    transition_block: Callable[..., tf.Tensor] = Field()
-    stem_filters: int = Field(64)
+def blurpool_initializer(shape, dtype=None):
```
leonoverweel, line 12: Can you add a docstring + reference here?

leonoverweel (approved):
Thanks for adding this, just one small request.

leonoverweel:
Should we do a Zoo release once this is merged and propagate these new numbers to the docs as well?

lgeiger:
> Should we do a Zoo release once this is merged and propagate these new numbers to the docs as well?

Yes, I think we should do a major release and update the docs

PR #259, opened 2020-11-20
Make flatten in global pool optional

jneeven:
For internal use

### DIFF
--- a/larq_zoo/core/utils.py
+++ b/larq_zoo/core/utils.py
@@ -103,7 +103,10 @@ def get_input_layer(input_shape, input_tensor):
 
 
 def global_pool(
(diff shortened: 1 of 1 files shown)


jneeven:
I'm not sure what's going on with the failing tests, I don't think that's related to my changes?

lgeiger (approved):
Very nice!

lgeiger:
> I'm not sure what's going on with the failing tests, I don't think that's related to my changes?

I don't think so, it looks like it only fails in TF 2.0 and 2.1 which is strange. Maybe a dependency changed?

lgeiger:
Closing this for now, as we do not need it internally anymore.

PR #264, opened 2020-12-02
Pin zookeeper@1.0.4 to fix CI

lgeiger:
[CI on master currently fails](https://github.com/larq/zoo/runs/1487893011), looks like https://github.com/larq/zookeeper/pull/206 broke this since downgrading to `zookeeper@1.0.4` seems to fix the issue.


jneeven:
@AdamHillier looks like an error is thrown during the new `__repr__` function?

PR #320, opened 2021-05-26
Use 1x1 convolutions to improve BinaryAlexNet latency on Compute Engine

lgeiger:
This PR updates `BinaryAlexNet` to use 1x1 convolutions in the classification layer which allows correct conversion to LCE.
This brings the official benchmarks inline with the numbers we quote in the LCE paper. This has a big impact since the classification layers are actually very large.

This code has been lying around on my system for a long time now, so I briefly spent a few minutes converting the weights which was just a matter of reshaping the weights of the last four classification layers.

### DIFF
--- a/larq_zoo/literature/binary_alex_net.py
+++ b/larq_zoo/literature/binary_alex_net.py
@@ -43,12 +43,12 @@ class BinaryAlexNetFactory(ModelFactory):
         )(x)
         if pool:
             x = tf.keras.layers.MaxPool2D(pool_size=3, strides=2)(x)
-        x = tf.keras.layers.BatchNormalization(scale=False, momentum=0.9)(x)
-        return x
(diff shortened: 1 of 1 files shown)


larq_zoo/literature/binary_alex_net.py
```diff
@@ -91,9 +92,9 @@ def build(self) -> tf.keras.models.Model:
             if self.include_top:
                 weights_path = utils.download_pretrained_model(
                     model="binary_alexnet",
-                    version="v0.2.0",
+                    version="v0.3.0",
                     file="binary_alexnet_weights.h5",
```
lgeiger, line 96: These weights have been manually reshaped from previous model so that Keras can correctly reload them.

jneeven (approved):

PR #331, opened 2021-07-05
Add TF 2.6.0rc0 to CI

lgeiger:


lgeiger:
This fix is a bit subtle since TF switched to `keras-team/keras` over `tensorflow/python/keras`. However, the old code still exists. Checkout https://github.com/tensorflow/tensorflow/issues/50613 for more information.

AdamHillier (approved):

PR #444, opened 2023-07-04
Update dependencies and deprecate Python 3.7 support

CNugteren:
This updates all the dependencies (and thus closes the other 7 open PRs) and drops Python 3.7 support (to satisfy the requirements of the dependencies).


setup.py
```diff
@@ -20,24 +20,27 @@ def readme():
     install_requires=[
         "numpy>=1.15.0",
         "packaging>=19",
-        "larq>=0.9.2,<0.14.0",
-        "zookeeper>=1.0.0,<1.4.0",
+        "larq>=0.9.2,<0.13.2",
```
CNugteren, line 23: The latest 0.13.2 caused tests to fail with:
```
TypeError: xnor_weight_scale() missing 1 required positional argument: 'x'
```
Probably related to some TF 2.13 compatibility changes in https://github.com/larq/larq/pull/831/files. So for now I pinned Larq to not use the latest.

lgeiger: Are you sure this is caused by the changes in https://github.com/larq/larq/pull/831, I am not sure what would actually cause this issue here 🤔

CNugteren, line 23: Let's try with the latest Larq instead:
```suggestion
        "larq>=0.9.2",
```

CNugteren: I've now reverted this change (and only this change) and we now have failing unittests.

lgeiger: Hmm interesting, I can take a look at it at some point. For now I agree, let's pin larq.

lgeiger (approved):

